PR #9, opened 2015-10-31
sendalerts concurrently

diwu1989:
this will allow alerts to be sent in parallel

i think you're going to have to have something like this in place to scale better

the __init__ files are necessary for the management commands to be picked up as modules

### DIFF
--- a/hc/api/management/commands/sendalerts.py
+++ b/hc/api/management/commands/sendalerts.py
@@ -2,12 +2,14 @@ import logging
 import sys
 import time
 
+from concurrent.futures import ThreadPoolExecutor
 from django.core.management.base import BaseCommand
+from django.db import connection
 from django.db.models import Q
 from django.utils import timezone
-
 from hc.api.models import Check
 
+executor = ThreadPoolExecutor(max_workers=10)
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,9 +1,10 @@
-Django==1.8.2
-django_compressor
 django-appconf
 django-ses-backend
-psycopg2==2.6
+Django==1.8.2
+django_compressor
 djmail
+futures
```
diwu1989: i alpha-sorted these, futures is the backport from Python3's concurrent package to Python2

also, you should consider bumping up the Django and psycopg2 versions

hc/api/management/commands/sendalerts.py
```diff
@@ -2,12 +2,14 @@
...
 
+from concurrent.futures import ThreadPoolExecutor
 from django.core.management.base import BaseCommand
+from django.db import connection
 from django.db.models import Q
 from django.utils import timezone
-
 from hc.api.models import Check
 
+executor = ThreadPoolExecutor(max_workers=10)
```
diwu1989: send up to 10 different alerts at the same time

hc/api/models.py
```diff
@@ -143,7 +143,7 @@ def notify(self, check):
                 "icon_url": "https://healthchecks.io/static/img/logo@2x.png"
             }
 
-            r = requests.post(self.value, json=payload)
+            r = requests.post(self.value, json=payload, timeout=5)
```
diwu1989: the webhook request has a timeout of 5
these other external API calls should also have the same timeout to prevent accidental stalls

diwu1989:
looks like i need to add unit test for coverage

cuu508:
Thanks for the pull request!

A couple of questions/notes:

    checks = list(query.iterator())

I had not used queryset.iterator() before. From reading on it, looks like it helps with memory usage when processing large number of items. But since they are put in a list anyway, is there a benefit of doing this, as opposed to just "list(query)"?

    for future in [executor.submit(handle_one, check) for check in checks]:
        future.result()

I'm new to futures too. Looks like the following would work too and be shorter?

    executor.map(handle_one, checks)

diwu1989:
`executor.map` should work too if we `list` the result because it returns an iterator,

it depends on whether you like the looks of 
```py
list(executor.map(handle_one, checks))
vs
for ....:
```

diwu1989:
if you `list` a `queryset`, the queryset itself still keeps an internal cache of the objects, see https://github.com/django/django/blob/master/django/db/models/query.py#L258

using iterator even when listing saves some additional cost, I don't mind taking out the .iterator() because its not a bottleneck

I'll find time later to add some good tests for this PR

cuu508:
Ah, I missed the fact that executor.map returns iterator not a list. In that case an explicit for loop, like you have, makes sense, let's keep that.

diwu1989:
thanks!

PR #11, opened 2015-11-16
make welcome code logic resilient

diwu1989:
if the temporary code ever disappears, you don't want the user to get stuck on the welcome screen
with tests


cuu508:
Thanks, this makes sense as soon as you start cleaning up old anonymous checks. I haven't done that so far but will have to, eventually.

diwu1989:
yeah, i implemented the anon cleanup, and then hit this bug

PR #12, opened 2015-11-17
Add PushOver integration

Schnouki:
This adds PushOver integration.

[PushOver](https://pushover.net) is a service that allows to push notifications to Android and iOS devices (and desktop computers through a web browser addon) using a simple API.

PO "applications" have a free quota of 7,500 messages per month. So the best way to integrate PO is to let the user create an application (it's free and takes less than 2 minutes), and then use this to send notifications.
Because of this, there are at least 2 things to store for the integration to work: the app API token, and the user key. I also like to configure the priority of messages sent with PushOver (an integer between -2 and 2).
In order to store these 3 values in the `Check.value` field, they are stored as pipe-separated values inside a string. It's a little hacky, but works fine.

Sample notification on my phone

Everything seems to work find; however please tell me if anything needs to be improved before you can merge it. (Or if you just don't like it and don't want to merge it -- that's fine, I mostly did it for fun :smiley:)


cuu508:
Thanks a lot for contributing!

I'll give it a go as soon as I get a chance, but at initial glance this integration looks fine, in line with the existing ones.

cuu508:
Hi Thomas,

tried this out–the integration page looks good and the integration itself works for me as well. Cool!

One concern I have: when registering a PushOver application, the registration form states:
"Each website, service, application, plugin, etc. may only be registered once and each application is currently limited to 7,500 messages per month."

I had a look at how StatusCake and Uptime Robot integrates with PushOver. Both of them only ask for User Key, and presumably use a site-wide API token. I think healthchecks.io should do the same. 

Would you mind moving pushover API token from "Add Integration" form to settings.py? In settings.py it could be set to None, and the actual API token would be in local_settings.py, which is not in version control. 

Other than that, I'd be happy to merge!

cuu508:
Asked PushOver support about multiple API keys, got this answer:

> If you are hosting this as a SaaS and not something the end user
> installs, you should have one API token that you register and use
> for all notifications.

> We also recommend using user subscriptions for obtaining user
> Pushover keys instead of having them copy/paste things.

Schnouki:
Sure thing. I did it like this because "free" applications are limited to 7.5k messages per month and I'm not sure how many could be sent by healthchecks.io.

So I moved to a site-wide API token in `settings.py` and using the subscription API, which is much nicer and simpler :) I'll push that very soon.

Schnouki:
Oops, the tests passed on my laptop since I have a valid API token in my `local_settings.py`... I'll fix this.

cuu508:
Looks good, thanks for contributing!

PR #14, opened 2015-11-28
redirect already logged in user

diwu1989:
If the user is already logged in, and then click the login email again, they get hit with this:
image

The reason is that their login token has been reset already, but they do have a valid session at the moment.

Correct behavior is to redirect the user to the checks screen instead.

and a small perf change that I saw while reading through the code

### DIFF
--- a/hc/front/views.py
+++ b/hc/front/views.py
@@ -162,8 +162,9 @@ def log(request, code):
     # Now go through pings, calculate time gaps, and decorate
     # the pings list for convenient use in template
     wrapped = []
+    now = timezone.now()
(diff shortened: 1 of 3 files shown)


hc/front/views.py
```diff
@@ -162,8 +162,9 @@ def log(request, code):
     # Now go through pings, calculate time gaps, and decorate
     # the pings list for convenient use in template
     wrapped = []
+    now = timezone.now()
```
diwu1989: time should be a fixed point in time, vs changing per iteration

cuu508:
Looks good, thanks!

PR #15, opened 2015-12-05
fix and optimize checks channel view

diwu1989:
- when u compare `channel.user` to `request.user`, it will issue a 2nd SQL query to fetch the related user object, do id comparison instead to save the extra query
- does the `channels.checks = new_checks` line really work as intended previous? I need to add unit test to verify
- grabbing `channel.checks.all()` and then grabbing only the `code` is not efficient compared to a `values_list`

### DIFF
--- a/hc/front/views.py
+++ b/hc/front/views.py
@@ -200,7 +200,7 @@ def channels(request):
             channel = Channel.objects.get(code=code)
         except Channel.DoesNotExist:
             return HttpResponseBadRequest()
-        if channel.user != request.user:
+        if channel.user_id != request.user.id:
             return HttpResponseForbidden()
 
         new_checks = []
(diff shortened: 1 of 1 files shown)


hc/front/views.py
```diff
@@ -211,11 +211,12 @@ def channels(request):
                     check = Check.objects.get(code=code)
                 except Check.DoesNotExist:
                     return HttpResponseBadRequest()
-                if check.user != request.user:
+                if check.user_id != request.user.id:
                     return HttpResponseForbidden()
                 new_checks.append(check)
 
-        channel.checks = new_checks
```
diwu1989: does django really do what u think it does when u do this kind of direct assignment?

cuu508: It's a m2m relation. I think it does:
https://docs.djangoproject.com/en/1.8/topics/db/examples/many_to_many/
"Relation sets can be assigned. Assignment clears any existing set members"

diwu1989:
hey, thanks for clarifying, i didn't get enough free time to circle back but glad to see u fixed it back on master

PR #20, opened 2015-12-19
check token need to handle invalid user & invalid token correctly

diwu1989:
Try going to this URL:
https://healthchecks.io/accounts/check_token/invalid_user/invalid_token/

You'll get hit with the login page, and u can enter email all day long and it wouldn't work

Correct solution is to:
- upon bad token or bad user, redirect over to login page
- set a `bad_link` sentinel in the cookie for the login page to render the special invalid login message
- clear the `bad_link` sentinel from session

with unit tests included

### DIFF
--- a/hc/accounts/tests/test_check_token.py
+++ b/hc/accounts/tests/test_check_token.py
@@ -1,15 +1,16 @@
 from django.contrib.auth.models import User
+from django.core.urlresolvers import reverse
 from django.test import TestCase
 
 
 class CheckTokenTestCase(TestCase):
 
     def setUp(self):
-      super(CheckTokenTestCase, self).setUp()
+        super(CheckTokenTestCase, self).setUp()
 
-      self.alice = User(username="alice")
-      self.alice.set_password("secret-token")
-      self.alice.save()
+        self.alice = User(username="alice")
+        self.alice.set_password("secret-token")
+        self.alice.save()
 
     def test_it_redirects(self):
         r = self.client.get("/accounts/check_token/alice/secret-token/")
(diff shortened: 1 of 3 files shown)


hc/accounts/tests/test_check_token.py
```diff
@@ -26,3 +27,10 @@ def test_it_redirects_already_logged_in(self):
         # Login again, when already authenticated
         r = self.client.get("/accounts/check_token/alice/secret-token/")
         assert r.status_code == 302
+
+    def test_it_redirects_bad_login(self):
+        # Login with a bad token
+        r = self.client.get("/accounts/check_token/alice/invalid-token/")
+        assert r.status_code == 302
+        assert r.url.endswith(reverse("hc-login"))
```
diwu1989: make sure we're redirecting to the right place

hc/accounts/tests/test_login.py
```diff
@@ -29,3 +29,8 @@ def test_it_sends_link(self):
         # And check should be associated with the new user
         check_again = Check.objects.get(code=check.code)
         assert check_again.user
+
+    def test_it_pops_bad_link_from_session(self):
+        self.client.session["bad_link"] = True
+        self.client.get("/accounts/login/")
+        assert "bad_link" not in self.client.session
```
diwu1989: make sure the user doesn't get stuck forever with the bad link login form

hc/accounts/views.py
```diff
@@ -110,8 +111,8 @@ def check_token(request, username, token):
 
             return redirect("hc-checks")
 
-    ctx = {"bad_link": True}
-    return render(request, "accounts/login.html", ctx)
```
diwu1989: dont render the login template, move user over to the login route instead

cuu508:
Looks good, thanks!

PR #22, opened 2015-12-19
fixed 50% width for the location column

diwu1989:
give the check location column a fixed 50% width so that when you switch between URL and Email, the column sizing doesn't change
image
image

### DIFF
--- a/static/css/my_checks_desktop.css
+++ b/static/css/my_checks_desktop.css
@@ -17,6 +17,10 @@ table.table tr > th.th-name {
     padding-left: 21px;
 }
 
+table.table tr > th.th-location {
+    width: 50%;
+}
+
 #checks-table .indicator-cell {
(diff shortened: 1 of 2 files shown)


cuu508:
Thanks for looking at this. The resizing when clicking on URL and Email is indeed quite annoying.
Problem with fixed 50% is that it might sometimes be not enough and sometimes too much space allocated for that column. I just commited a fix which 

* allocates just enough space for URL to fit. I'm assuming email address will be shorter than URL. When switching, it sets "height: 0" to keep the allocated space, but avoid resizing and flickering.
* makes the URLs and email addresses selectable with a single click.

diwu1989:
cool, just reverted mine on my fork and merged with ur upstream/master
looks great, glad to see this annoyance fixed finally

PR #25, opened 2015-12-31
n_pings is unnecessary

diwu1989:
you don't need to do the book-keeping of n-pings yourself
it's just a count of the number of pings for that check, and can be expressed in a single SQL
the code for pruning the pings and re-setting the n-pings value is also vulnerable to race condition


hc/api/management/commands/prunepings.py
```diff
@@ -9,18 +10,20 @@ class Command(BaseCommand):
...
             Profile.objects.for_user(user)
 
-        # Select checks having n_ping greater than the limit in user profile
+        # Select checks having more pings than the limit in user profile
         checks = Check.objects
-        checks = checks.annotate(limit=F("user__profile__ping_log_limit"))
-        checks = checks.filter(n_pings__gt=F("limit"))
+        checks = checks.annotate(
+            limit=F("user__profile__ping_log_limit"),
+            count_pings=Count("ping"))
```
diwu1989: you can annotate the `Count` of the pings for the check like this

hc/api/management/commands/prunepings.py
```diff
@@ -9,18 +10,20 @@ class Command(BaseCommand):
...
+        checks = checks.filter(count_pings__gt=F("limit"))
```
diwu1989: and then you can filter based on that `count`
this translates into a `HAVING` clause in SQL if you want to look at the underlying query

hc/api/management/commands/prunepings.py
```diff
@@ -9,18 +10,20 @@ class Command(BaseCommand):
...
+        checks = checks.select_related("user").order_by("user", "id")
```
diwu1989: you need the `select_related` because in your for loop, you are logging the user's email

hc/api/models.py
```diff
@@ -112,13 +111,10 @@ def prune_pings(self, keep_limit):
         if len(cutoff) == 0:
             return 0
 
-        cutoff_date = cutoff[0].created
-        q = Ping.objects.filter(owner=self, created__lte=cutoff_date)
+        cutoff_id = cutoff[0].id
```
diwu1989: using `created` a timestamp is not safe, two checks may have the exact same `created` time, but `id` is guaranteed to be unique and monotonically increasing

cuu508:
Counting pings on the fly, without the `n_pings` field, does have several upsides: there's less stuff in database, less code to keep it up-to-date and no chance of that number diverging from the correct value.

My reasoning behind `n_pings` field is related to performance. Without `n_pings` field, on each `prunepings` run I would need to count pings for each check, to see if it needs pruning. For, say, one thousand checks, that's one thousand fairly expensive SQL queries. And I want to run `prunepings` often, so there's not much accumulated pings between runs, and each run is quick.

The `prunepings` command I had initially was counting pings on the fly. On the production database, on a fast machine, each no-op run took 15 seconds to complete. With the `n_pings` field, each no-op run takes less than a second (one scan of `api_check` table).

cuu508:
PS good points about `select_related` and using `cutoff_id` instead of `cutoff_date` -- I'll apply these improvements!

diwu1989:
`For, say, one thousand checks, that's one thousand fairly expensive SQL queries.`
it's not, its a single query, that's the whole point of this
```py
checks = checks.annotate(
         limit=F("user__profile__ping_log_limit"),
         count_pings=Count("ping"))
```

diwu1989:
take a backup of your database and then run my version against it, each no-op run should not take 15s

cuu508:
Ah, you're right, it's one query. Mixed up with a previous experiment where I was doing N calls.

So it's 1 query but it does need to do sequential scan on `api_ping` table:

    # explain SELECT "api_check"."id", "api_check"."name", "api_check"."tags", "api_check"."code", "api_check"."user_id", "api_check"."created", "api_check"."timeout", "api_check"."grace", "api_check"."n_pings", "api_check"."last_ping", "api_check"."alert_after", "api_check"."status", COUNT("api_ping"."id") AS "count_pings", "accounts_profile"."ping_log_limit" AS "limit", "auth_user"."id", "auth_user"."password", "auth_user"."last_login", "auth_user"."is_superuser", "auth_user"."username", "auth_user"."first_name", "auth_user"."last_name", "auth_user"."email", "auth_user"."is_staff", "auth_user"."is_active", "auth_user"."date_joined" FROM "api_check" LEFT OUTER JOIN "api_ping" ON ("api_check"."id" = "api_ping"."owner_id") LEFT OUTER JOIN "auth_user" ON ("api_check"."user_id" = "auth_user"."id") LEFT OUTER JOIN "accounts_profile" ON ("auth_user"."id" = "accounts_profile"."user_id") GROUP BY "api_check"."id", "accounts_profile"."ping_log_limit", "auth_user"."id" HAVING COUNT("api_ping"."id") > ("accounts_profile"."ping_log_limit")
    hcg-# ;
                                                 QUERY PLAN                                             
    ----------------------------------------------------------------------------------------------------
     GroupAggregate  (cost=210783.48..212235.81 rows=52812 width=235)
       Group Key: api_check.id, accounts_profile.ping_log_limit, auth_user.id
       Filter: (count(api_ping.id) > accounts_profile.ping_log_limit)
       ->  Sort  (cost=210783.48..210915.51 rows=52812 width=235)
             Sort Key: api_check.id, accounts_profile.ping_log_limit, auth_user.id
             ->  Hash Left Join  (cost=292.10..200683.76 rows=52812 width=235)
                   Hash Cond: (auth_user.id = accounts_profile.user_id)
                   ->  Hash Left Join  (cost=277.42..199942.91 rows=52812 width=231)
                         Hash Cond: (api_check.user_id = auth_user.id)
                         ->  Hash Right Join  (cost=256.73..199197.01 rows=52812 width=109)
                               Hash Cond: (api_ping.owner_id = api_check.id)
                               ->  Seq Scan on api_ping  (cost=0.00..198214.12 rows=52812 width=8)
                               ->  Hash  (cost=242.99..242.99 rows=1099 width=105)
                                     ->  Seq Scan on api_check  (cost=0.00..242.99 rows=1099 width=105)
                         ->  Hash  (cost=14.75..14.75 rows=475 width=122)
                               ->  Seq Scan on auth_user  (cost=0.00..14.75 rows=475 width=122)
                   ->  Hash  (cost=8.75..8.75 rows=475 width=8)
                         ->  Seq Scan on accounts_profile  (cost=0.00..8.75 rows=475 width=8)
    (18 rows)

`api_ping` can conceivably grow into millions of rows, and it wouldn't be great to regularly do sequential scans on it then.

diwu1989:
yes there's no way around reading through the ping table to tally up all the counts

even if api_pings grow into millions of rows, how often do you need this management command to be run? I have it configured for hourly on my server instance right now, and I think even for millions of pings, it should still be just a few seconds to read through them.

Looks like you have 52k rows in the pings table right now, what's the current timing?

cuu508:
I'm closing this PR: `n_pings` now stores total number of pings the check has ever received, not the number of pings currently stored in database.

A side note–I'll want to run the `prunepings` command fairly often, so each run doesn't have too much impact. If, for example, a live server receives 5 pings per second, over an hour this accumulates to 18K pings. Haven't decided on frequency yet, but it's probably going to be every 10 minutes or so.

diwu1989:
k, i see,

I think if you ultimately want to scale up Healthchecks beyond a hobby project, then you'll probably have to introduce Celery into the stack and do proper background task distribution. The current way of sending notifications definitely isn't scalable to a real service offering, not even with the multi-threaded code

cuu508:
Yes, a real queue may be needed as the service scales up. The notifications are much less frequent than pings though, and it might be long time until a single "sendalerts" process cannot keep up.

I've been thinking and experimenting with HA stuff quite a bit recently. One of the pieces of the puzzle is being able to run two "sendalerts" worker processes concurrently. So if one goes away, the other one picks up the slack. Obviously they must not send the notifications twice so there needs to be some kind of a locking mechanism. 

Assume the database (PostgreSQL) is beefy, can take the load, has automatic failover, etc. Assume there's no separate message broker. Here's what I'm considering:

1. Assign an unique name to each worker process running the "sendalerts" command
2. Add two new fields to `api_check` table: `alert_worker` (varchar) and `alert_date` (date). 
3. Each worker polls the `api_check` table and looks for alerts that need to be sent–same as now. But when it finds one:

- it updates `alert_worker` and `alert_date` columns with worker's name and current date, and commits
- it reads both values back
- if either of the values has changed, it does nothing and goes back to polling for more work
- if both values are unchanged, it goes ahead and sends the alert
- after the alert is sent, it clears out `alert_worker` and `alert_date` fields

My thinking is, if the database supports at least "read commited" isolation level, this should prevent multiple workers from sending duplicate alerts. The `alert_worker` field is effectively an application-level lock, and the `alert_date` can be used for calculating its expiry date (in case a worker dies before it blanks out the the two fields).

Alternatively, with PostgreSQL, the advisory locks could be used, and would likely perform better. But then MySQL would still need a separate mechanism like the above.

@diwu1989 thoughts on this?

PR #27, opened 2016-01-05
fix tests and optimize authentication and fix some unclosed markup

diwu1989:
later, the test case should be refactored out into a custom base test case that creates the `alice` user so that there's not so much redundant code


cuu508:
Looks good, thanks!

PR #30, opened 2016-01-18
use basetestcase for most test cases that require predefined user

diwu1989:
should not be duplicating so much code for the base test case setup

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,4 @@ addons:
 script:
   - coverage run --source=hc manage.py test
 after_success: coveralls
+cache: pip
--- a/hc/front/tests/test_add_channel.py
+++ b/hc/front/tests/test_add_channel.py
@@ -1,19 +1,11 @@
-from django.conf import settings
-from django.contrib.auth.models import User
-from django.test import TestCase
+from django.test.utils import override_settings
+
 from hc.api.models import Channel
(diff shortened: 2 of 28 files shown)


hc/front/tests/test_add_channel.py
```diff
@@ -1,19 +1,11 @@
...
+from hc.test import BaseTestCase
 
-    def setUp(self):
-        super(AddChannelTestCase, self).setUp()
-        self.alice = User(username="alice", email="alice@example.org")
-        self.alice.set_password("password")
-        self.alice.save()
 
-        settings.PUSHOVER_API_TOKEN = "bogus_token"
-        settings.PUSHOVER_SUBSCRIPTION_URL = "bogus_url"
```
diwu1989: changing settings should be done via the `@override_settings` decorator instead

.travis.yml
```diff
@@ -14,3 +14,4 @@ addons:
 script:
   - coverage run --source=hc manage.py test
 after_success: coveralls
+cache: pip
```
diwu1989: should be caching the pip dependencies for faster builds, the actual test runs in a few seconds, but up to 2 mins are spent just re-installing stuff all the time

cuu508:
Thanks for this!
Also, thanks for the ensure_triggers testcase.

PR #31, opened 2016-01-19
dont include tests in coverage calculation

diwu1989:
I noticed that in coverall, the test files were counting towards code coverage
tests should be excluded from coverage information


diwu1989:
coverage information dropping is expected, this just means we should be writing unit tests to cover more of the real logic

PR #32, opened 2016-01-19
defer inserting the password field into the dom

diwu1989:
the password input field is hidden in the DOM, but that still means password managers will pick it up and insert into a password to use

if I have a password manager with multiple accounts + passwords, this can screw up the login experience

if the password input field is ONLY inserted into the DOM after the toggle has been checked, then the password managers will not be tricked.

I know this is an edge case with how browser password managers work, but the way I fixed it doesn't cause any interaction flow to existing users.

### DIFF
--- a/static/js/login.js
+++ b/static/js/login.js
@@ -1,7 +1,9 @@
 $(function () {
+    var passwordInput = $("#password-block input");
+    passwordInput.detach();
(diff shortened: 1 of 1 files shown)


cuu508:
Thanks, good stuff, the password manager + the hidden password field has annoyed myself as well.

PR #33, opened 2016-01-19
forwarded for and forwarded protocol

diwu1989:
when an incoming http request comes through a proxy, it will get the header `HTTP_X_FORWARDED_FOR` and `HTTP_X_FORWARDED_PROTO`

see protocol information https://en.wikipedia.org/wiki/X-Forwarded-For

The `ping` API should also hard enforce that its response not be cached so that any proxy in between don't screw it up.

### DIFF
--- a/hc/api/tests/test_ping.py
+++ b/hc/api/tests/test_ping.py
@@ -5,33 +5,28 @@ from hc.api.models import Check, Ping
 
 class PingTestCase(TestCase):
 
-    def test_it_works(self):
-        check = Check()
-        check.save()
+    def setUp(self):
+        super(PingTestCase, self).setUp()
+        self.check = Check.objects.create()
 
-        r = self.client.get("/ping/%s/" % check.code)
+    def test_it_works(self):
+        r = self.client.get("/ping/%s/" % self.check.code)
         assert r.status_code == 200
 
-        same_check = Check.objects.get(code=check.code)
+        same_check = Check.objects.get(code=self.check.code)
         assert same_check.status == "up"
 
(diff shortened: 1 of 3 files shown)


hc/api/tests/test_ping.py
```diff
@@ -44,22 +39,51 @@ def test_it_handles_120_char_ua(self):
...
+
+    def test_it_reads_forwarded_ip(self):
+        ip = "1.1.1.1"
+        r = self.client.get("/ping/%s/" % self.check.code,
+                            HTTP_X_FORWARDED_FOR=ip)
+        ping = Ping.objects.latest("id")
+        assert r.status_code == 200
+        assert ping.remote_addr == "1.1.1.1"
+
+        ip = "1.1.1.1, 2.2.2.2"
```
diwu1989: if the incoming request passed through **multiple** proxies on its way, this will be a list of IPs and the first one is the originating source

cuu508:
I think we should also drop the code for `X-Real-IP` and `X-Scheme` headers. The `X-Forwarded-For` appears to be way more common.

Side note: I think we can also assume that in production the server will always be run behind some kind of a reverse proxy or load balancer. The only time we use `REMOTE_ADDR` value would be in a development environment.

diwu1989:
yeah i'm running my deployment of healthcheck behind nginx

cuu508:
Me too, looks like it would be easy enough to switch nginx configuration from `X-Real-IP` to `X-Forwarded-For`. And that would play well with things like Amazon ELB.

PR #34, opened 2016-01-19
remove channel should always redirect even if removal fails

diwu1989:
removing the channel may be a slow operation and if the user refresh in the middle of the request, Chrome will make a 2nd POST request, which will end up erroring

I think it's better to always redirect the user to the channel list regardless of whether the removal UUID was valid or not, better than showing an error page if the user sends two removals simultaneously

### DIFF
--- a/hc/front/views.py
+++ b/hc/front/views.py
@@ -336,11 +336,11 @@ def verify_email(request, code, token):
 def remove_channel(request, code):
     assert request.method == "POST"
 
-    channel = get_object_or_404(Channel, code=code)
-    if channel.user != request.user:
-        return HttpResponseForbidden()
-
-    channel.delete()
+    channel = Channel.objects.filter(code=code).first()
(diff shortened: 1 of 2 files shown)


hc/front/views.py
```diff
@@ -336,11 +336,11 @@ def verify_email(request, code, token):
 def remove_channel(request, code):
     assert request.method == "POST"
 
-    channel = get_object_or_404(Channel, code=code)
-    if channel.user != request.user:
-        return HttpResponseForbidden()
-
-    channel.delete()
+    channel = Channel.objects.filter(code=code).first()
```
cuu508: Would be a good idea to add a code comment here, explaining why `filter().first()` is used instead of `get_object_or_404()`

diwu1989: yeah I'll ammend, thanks for the suggestion

PR #36, opened 2016-01-22
Add healthcheck UA to inform server about identity

zonito:
Adding "healthcheck.io" as User-Agent to inform webhook handler server about request is specific from healthcheck.


cuu508:
Makes sense to do this, thanks!

zonito:
Do update, once you deploy new code to server including this changes.

PR #40, opened 2016-02-03
Add VictorOps integration

ssunagari:
We are evaluating a move from PagerDuty to VictorOps, and also looking at healthchecks.io, so having an integration was desirable... this is my first interaction with Python and Django so the effort consisted mostly of copy-paste-edit wherever obvious.  Hope it's usable :-)  Did functional testing on local machine with `./manage.py runserver` to add an integration from the UI and `./manage.py sendalerts`.


cuu508:
Thanks for doing this! I'll make few small changes to JSON payload that is sent to VictorOps, but other than that it looks good!

PR #50, opened 2016-04-04
Implement GET /api/v1/checks/ for listing existing checks

ssunagari:
Another sort-of-rough-draft pull request for allowing an API client to obtain a list of existing checks.  It's missing the list of channels and probably some bullet-proofing.

### DIFF
--- a/hc/api/tests/test_list_checks.py
+++ b/hc/api/tests/test_list_checks.py
@@ -1,0 +1,48 @@
+import json
+from datetime import timedelta as td
+
+from hc.api.models import Check, User
+from hc.test import BaseTestCase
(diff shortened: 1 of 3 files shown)


cuu508:
Looks good!

In the JSON response, I would rename `url` key to `ping_url` – so it is consistent with the resonse from "POST /api/v1/checks/". 

Out of interest, what would you be using this API call for?

ssunagari:
We're looking at incorporating this into our Terraform/Ansible
provisioning to create a health check for each machine as it's brought
up.  Since this will key on hostname, it seemed useful to see whether
there was an existing check for that host since we may destroy a
machine and reprovision it later.

PR #60, opened 2016-05-23
fix some markup and be more explicit about date time format

diwu1989:
image
and some other markup fixes along the way


templates/front/my_checks.html
```diff
@@ -43,13 +43,12 @@
     </div>
 </div>
 <div class="row">
-    <div class="col-sm-12"></div>
```
diwu1989: i think u meant for this col to wrap the form

templates/front/my_checks.html
```diff
@@ -177,7 +176,7 @@ <h4 class="update-timeout-title">Name and Tags</h4>
             {% csrf_token %}
             <div class="modal-content">
                 <div class="modal-header">
-                    <button type="button" class="close" data-dismiss="modal">&times;</span></button>
```
diwu1989: span closing tag typo

templates/front/my_checks_desktop.html
```diff
@@ -77,7 +77,7 @@
         {% if check.last_ping %}
             <span
                 data-toggle="tooltip"
-                title="{{ check.last_ping }} GMT">
+                title="{{ check.last_ping|date:'N j, Y, P e' }}">
```
diwu1989: taken from https://docs.djangoproject.com/en/1.9/ref/settings/#std:setting-DATETIME_FORMAT
with `e` added at the end to make sure we are explicit about which timezone

PR #72, opened 2016-07-15
exposes the status field over the rest api

tillkahlbrock:
I would like to be able to additionally fetch the status of a check through the rest api.

### DIFF
--- a/hc/api/models.py
+++ b/hc/api/models.py
@@ -114,7 +114,8 @@ class Check(models.Model):
             "tags": self.tags,
             "timeout": int(self.timeout.total_seconds()),
             "grace": int(self.grace.total_seconds()),
-            "n_pings": self.n_pings
+            "n_pings": self.n_pings,
+            "status": self.get_status()
         }
 
         if self.last_ping:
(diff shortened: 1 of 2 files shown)


hc/api/models.py
```diff
@@ -114,7 +114,8 @@ def to_dict(self):
             "tags": self.tags,
             "timeout": int(self.timeout.total_seconds()),
             "grace": int(self.grace.total_seconds()),
-            "n_pings": self.n_pings
+            "n_pings": self.n_pings,
+            "status": self.status
```
cuu508: Use self.get_status() to compute the correct status on the fly. self.status may be out of date if sendalerts is not running or is swamped with work. Also, should probably include a "in_grace_period" flag: 

    "status": "grace" if self.in_grace_period() else self.get_status()

tillkahlbrock: Your are definitely right about `get_status()`. I will change that. 

But I'am not sure about returning `grace` as status, since it is not really a status of a check according to https://github.com/healthchecks/healthchecks/blob/master/hc/api/models.py#L16-L21.

cuu508: Yes, a cleaner option would be to have a separate flag for grace status. And we can add it when there's actually a need for it.

Issue #82, opened 2016-09-21
API request: optionally unique check names


foozmeat:
It would simplify automated deployments if new checks could be created with a flag specifying that they should be unique. If the check with the requested name already exists, return its info with a `200` code, otherwise return a `201 CREATED`.

cuu508:
Fixed in #85

PR #85, opened 2016-09-22
adds a unique parameter to the check creation API

foozmeat:
It only checks for name uniqueness.

for #82

### DIFF
--- a/hc/api/views.py
+++ b/hc/api/views.py
@@ -55,8 +55,20 @@ def checks(request):
         return JsonResponse(doc)
 
     elif request.method == "POST":
+
+        unique = bool(request.json.get("unique", False))
+        name = str(request.json.get("name", ""))
+
+        if unique:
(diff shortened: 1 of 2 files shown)


cuu508:
Hi @foozmeat , looks like useful feature, thanks for doing this!
Maybe instead of `unique=True` go with `unique=["name"]`. It's sort-of self-documenting, and a bit more future proof.
Also would be good to mention 200/201 status codes in the docs.

foozmeat:
so `unique` would specify which field you want to be unique in the request?

I can update the docs with the result codes.

cuu508:
Yes! Down the road we might want to support e.g. "require unique combination of name and tags" and express it as `unique=["name", "tags"]`.

foozmeat:
I just went ahead and implemented unique checks for name, tags, timeout, and grace.

Issue #79, opened 2016-09-01
Add SITE_NAME setting, use it in templates instead of the hardcoded "healthchecks.io"


cuu508:
From @brumle80:

> not critical, but the notification email should use the site url automatically instead of healthchecks.io if it's possible without manually editing all the templates :)

PR #86, opened 2016-09-22
adds a SITE_NAME setting that's used to label the site

foozmeat:
should mostly address #79 

I left alone any pages related to the healthckeck.io business like billing, privacy, and about.

One small change that should probably be made is to add a variable for a private site's slack app name so the slack docs will match what a user might see.


cuu508:
The `{% settings_value "SITE_NAME" %}` and `{% settings_value "SITE_ROOT" %}` repeat a lot, so I'm thinking it would be nice to have more concise templates with specialized versions: `{% site_name %}` and `{% site_root %}`.
What do you think?

With `SITE_ROOT` being in `settings.py`, can we change `getattr(settings, "SITE_ROOT")` with `settings.SITE_ROOT` or am I missing something? (and same with `SITE_NAME`)

foozmeat:
I'm fine with both. I can update my PR to reflect this.

cuu508:
Looks good, thanks!

Issue #88, opened 2016-09-28
Add "copy to clipboard" function to example code snippets


cuu508:
For the example code snippets (bash, python, ruby, etc.), it would be great to have a "copy" buttons next to each. 

The snippets are in several places on the site:

 * the welcome page for not-logged-in users
 * documentation
 * My Checks → gear icon → Usage Examples

The project is already using clipboard.js for ping addresses.

eddieosmanbasic:
Hello,

I'm new to contributing to open source projects, and using GitHub in general. I was wondering if this button needs to be written in Python considering the fact that the website uses Django?

cuu508:
Hi @eddieosmanbasic – no, this would be done entirely on client side, using JS. 
If you log into healthchecks.io and look at the list of your checks, there is a "copy" link next to each ping URL. It works, it uses clipboard.js under the hood. This issue is about extending the same functionality to more places on the site.

A quick note: clipboard.js and similar libraries do not work on OSX. There, we fall back to simply not showing the "copy" link/button.

cdax:
Hi @cuu508, I've been thinking about how this should look from a UI/UX perspective. There are cases where the same tab can have multiple independent code snippets. For example, have a look at the *Bash* tab:
<img width="654" alt="screen shot 2016-09-30 at 4 59 26 pm" src="https://cloud.githubusercontent.com/assets/7029291/18990434/6f8292aa-872f-11e6-9ba0-a22b769aefda.png">
here, you can't have a single floating "Copy" button for the whole tab since the `curl` snippet is independent of the `wget` snippet.

One solution might be to break down independent snippets into separate [Bootstrap panels](http://getbootstrap.com/components/#panels). I should be able to finish working on this over the weekend once we've reach some agreement around how this should look.

cuu508:
Yes, here, copying the whole snippet, comments and all, is not very useful.

I think I would prefer to still have a single "Bash" tab, but split this snippet in two parts: `front/snippets/curl.txt` and `front/snippets/wget.txt`. 

The comments could also be moved from the snippet to the parent template.

cuu508:
PR from @cdax received and merged.

PR #90, opened 2016-10-01
Adds 'copy to clipboard' function to example code snippets

cdax:
Adds copy-to-clipboard links to code snippets in the following places:
- the welcome page for not-logged-in users
- documentation
- My Checks → gear icon → Usage Examples

I'm using clipboard.js, and I'm setting the target text dynamically by parsing the corresponding pygments `.highlight` element, and ignoring all text within pygments's comment classes. This allows the snippet text files to have as many comments as we want while ensuring that they're never copied to the clipboard along with the actual code.

Implements #88

### DIFF
--- a/hc/front/management/commands/pygmentize.py
+++ b/hc/front/management/commands/pygmentize.py
@@ -27,10 +27,12 @@ class Command(BaseCommand):
             return
 
         # Invocation examples
-        _process("bash", lexers.BashLexer())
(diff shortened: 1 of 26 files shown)


cuu508:
Thanks! Removing comments in JS–clever!
I replaced the "copy" link with an icon, updated the JS code to inject "copy" buttons at runtime (so same markup doesn't need to be repeated over and over again), did some CSS tweaking, and merged.

PR #114, opened 2017-02-22
Heroku compatibility

iphoting:
For your consideration, some minor tweaks to the project such that it can be nicely deployed onto Heroku, configured to use Heroku add-ons: Postgresql and Mailgun (for outbound email).


iphoting:
Issues noted will be resolved first. Closing.

PR #117, opened 2017-03-11
Add --no-loop option to sendalerts management command.

iphoting:
Allow`sendalerts` to be run by an external scheduler, on a one-shot basis, exiting after all alerts have been sent.

### DIFF
--- a/hc/api/management/commands/sendalerts.py
+++ b/hc/api/management/commands/sendalerts.py
@@ -26,6 +26,15 @@ class Command(BaseCommand):
     help = 'Sends UP/DOWN email alerts'
     owned = Check.objects.filter(user__isnull=False)
 
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--no-loop',
+            action='store_false',
+            dest='loop',
+            default=True,
+            help='Do not keep running indefinitely in a 2 second wait loop',
+        )
+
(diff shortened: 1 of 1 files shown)


cuu508:
Hello, thanks for the PR!

A couple comments:

* In the `--no-loop` case, would it not make sense to keep calling `self.handle_one()` until it returns `False`? e.g., if there are 3 alerts to be sent, currently it will only send one per run. If the external scheduler calls `manage.py sendalerts --no-loop` every 5 minutes, it will take 5+5 minutes until all three are sent.

* For the sake of consistency, I would also change `sendreports` command to loop by default, and have a `--no-loop` flag for one-shot invocations.

iphoting:
> In the --no-loop case, would it not make sense to keep calling self.handle_one() until it returns False? e.g., if there are 3 alerts to be sent, currently it will only send one per run. If the external scheduler calls manage.py sendalerts --no-loop every 5 minutes, it will take 5+5 minutes until all three are sent.

Good point, I must have misread the code; will definitely fix this.

> For the sake of consistency, I would also change sendreports command to loop by default, and have a --no-loop flag for one-shot invocations.

I would have thought that `sendreports` would only need to fire once a day or week; having it loop by default would be a little too excessive?

PR #134, opened 2017-09-01
Tags querystring param to filter checks by tags

umitakkaya:
I thought filtering checks by tags through API would be nice to have.
I am not a python guru or django guru, so feedbacks to improve are welcome.

Update: Documentation is going to be updated in this PR as well.

### DIFF
--- a/hc/api/tests/test_list_checks.py
+++ b/hc/api/tests/test_list_checks.py
@@ -20,6 +20,7 @@ class ListChecksTestCase(BaseTestCase):
         self.a1.last_ping = self.now
         self.a1.n_pings = 1
         self.a1.status = "new"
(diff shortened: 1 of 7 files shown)


cuu508:
Hi @umitakkaya 

thanks--I think this would be an useful feature! A couple notes:

1. Results can be slightly inaccurate if tags have similar prefixes. For example, if there are tags `staging` and `staging-2`, then filtering by `staging` will match both. Check out `hc.api.views.badge` view for a simple workaround.
2. It would be great to also mention this feature in API docs.
3. Maybe rename `?tags` parameter to `?tag` as it accepts a single tag.

umitakkaya:
@cuu508 I guess now it's done :)

cuu508:
Looks good, thanks!

Issue #120, opened 2017-04-02
Headers & Auth for Webhook


tester22:
Would be good if it was possible to set basic auth and/or headers to outgoing webhooks.

cuu508:
Basic authentication you can already do like so:

    http://foo:bar@example.org

cuu508:
Webhooks now support custom request headers. 
Initial implementation by @someposer in PR #140

PR #140, opened 2017-11-03
Adding Content-Type header to Webhook integrations

someposer:
Adding Content-Type header to Webhook integrations to work correctly with services like [IFTTT Maker Webhooks](https://ifttt.com/maker_webhooks) which require a specific content type, like application/json.

If the content-type is not provided, the post data is not parsed by IFTTT, which prevents the ability to use variables like, `$NAME` and `$STATUS`

### DIFF
--- a/hc/api/models.py
+++ b/hc/api/models.py
@@ -322,6 +322,12 @@ class Channel(models.Model):
         return parts[2] if len(parts) > 2 else ""
 
     @property
+    def content_type(self):
+        assert self.kind == "webhook"
+        parts = self.value.split("\n")
+        return parts[3] if len(parts) > 3 else ""
+
+    @property
     def slack_team(self):
         assert self.kind == "slack"
         if not self.value.startswith("{"):
--- a/hc/api/models.py
+++ b/hc/api/models.py
@@ -304,28 +304,44 @@ class Channel(models.Model):
         return user_key, prio, PO_PRIORITIES[prio]
 
     @property
-    def value_down(self):
+    def url_down(self):
         assert self.kind == "webhook"
-        parts = self.value.split("\n")
-        return parts[0]
+        if not self.value.startswith("{"):
+            parts = self.value.split("\n")
+            return parts[0]
+
+        doc = json.loads(self.value)
+        return doc["url_down"]
+        
 
     @property
-    def value_up(self):
+    def url_up(self):
         assert self.kind == "webhook"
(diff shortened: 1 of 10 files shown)


cuu508:
Hi @someposer, thank you for the pull request!

It would be great to be able to specify the content type for webhooks. 

I see your code adds the 4th row in the Channel.value field. I think it would make sense to upgrade the Channel.value format so that in future there's place for arbitrary request headers (see issue #120). Maybe something similar as slack and hipchat integrations: Channel.value can be either 1 to 3 lines separated with `\n`, or a JSON string, for example:


    {"url_up": "...", "url_down": "...", "body": "...", "headers": ["Content-Type: application/json"]}

cuu508:
PS. Also, let's keep this PR focused on a single feature. It is OK if the coverage decreases slightly. The test cases for other areas are very welcome, but please submit them in separate PRs.

someposer:
I'll update the format to use JSON instead of a multiline value and clean up the PR.

someposer:
@cuu508 I've added support for arbitrary headers which should solve #120 and provide enough flexibility for what I originally intended for IFTTT.

cuu508:
@someposer thank you for the fixes. One more thing before I'm happy to merge this:

The `headers` structure in Channel.value is double-encoded JSON. Also, if I understand it correctly, it requires the end user to use JSON notation when specifying headers.

The UI for end users that I'm thinking of would be a flexible number of textbox pairs for entering header names and values. As an example, here's a screenshot from Postman, an app where users can craft HTTP requests:

image

That would take some work to implement. But we can initially support just this one header field, and add "arbitrary headers" feature later on. The form would be:

* URL for "down" events: [ ____ ]
* URL for "up" events: [ ____ ]
* POST data: [ ____ ]
* Content-Type header: [ ____ ]

When saving form data, the code would do `json.dumps({...., "headers": {"Content-Type": form.cleaned_data["contenttype"]}})

Would you be OK to update the PR to work like that?

someposer:
@cuu508 I've added an arbitrary headers UI similar to what you suggested. The webhook data is no longer double JSON encoded.

cuu508:
Thank you, from an initial look -- excellent work!
I'm bogged down with other work at the moment, but will review more carefully and merge as soon as I can.

cuu508:
I made a few tweaks to the UI and code, and have now deployed this on https://healthchecks.io
Thank you again for your work!

PR #142, opened 2017-11-28
Minor HTML edit on my_checks.html

fishnux:
I wanted to check the cron cheatsheet and got annoyed because it opened on the same tab, so I had to go back and open the window/modal to edit the cron period, and open the cheatsheet again, but this time on a new tab, so I could build my cron expression side-by-side.


cuu508:
Good idea, thanks!

PR #143, opened 2017-12-14
Creates a native PagerTree integration

armiiller:


templates/front/welcome.html
```diff
@@ -319,6 +319,13 @@ <h2 class="tour-title">A quick peek of what's inside:</h2>
...
+                </td>
+                <td>Open and resolve incidents in <a href="https://www.pagertreecom/">PagerTree</a>.</td>
```
cuu508: A missing dot in URL

cuu508:
Hello @armiiller, thank you for making this integration!

I wanted to try it out, so I signed up for a PagerTree account, added the integration and opened an incident. On PagerTree, I noticed the description is shown as a JSON document:

image

Is this as-intended? From a quick look I could not find documentation on what the webhook payload should look like. I'm assuming it's the same structure as PagerDuty?

Another thing, on the "Add Integration" page it would be great to have step-by-step instructions on what to do on the PagerTree site (how to get the webhook URL for pasting in healthchecks.io).

templates/front/welcome.html
armiiller: Good catch - I'll fix this

armiiller:
Hi @cuu508 - Absolutely! Really like this tool and its straight forward and simple design.

As for the JSON document showing, this is currently the intended functionality. We have a [task](https://pagertree.com/roadmap/2018/#enhanced-incident-view) in early 2018 to go through each of our integrations and add specific views with links to 3rd party urls and support other features from each integration. But as for now, that behavior is correct.

Also **do not** use the same format as PagerDuty. We write adapters for each third party integration.

I think you're right in regards to documentation on the add integration page, I'll add that. I debated this since we will also post documentation on the PagerTree website. But for ease of use to the user it would be nice to heave it inside of healthchecks, there in front, without having to go to another page for documentation. Look for the coming pull request.

Also, as a side note, super happy you signed up :) We haven't opened up sms or voice support to Latvia, but if that's something your interested in we could get that done for you.

armiiller:
@cuu508 - The latest push should have taken care of all the comments. Let me know if there is anything else. We've added our documentation [here](https://pagertree.com/knowledge-base/integration-healthchecks/).

Sincerely,
-Austin

cuu508:
Thank you for these updates, looks good!

PR #149, opened 2018-01-18
Added deterministic handling of sqlite database location

blarghmatey:
When running the migration command from outside of the application directory the sqlite database is created in the current working directory at the time of the command being executed. This commit updates the file path to be relative to the location of the settings file.


cuu508:
Hello @blarghmatey,

yes, makes sense to store the database file relative to `settings.py`, not relative to current working directory.

Check out `BASE_DIR` in In `settings.py`, line 16. The change could be a simple

        'NAME': BASE_DIR + '/hc.sqlite',

PR #163, opened 2018-03-19
Update docs to reflect new maximum timeout value

brianstorti:
Based on this change: https://github.com/healthchecks/healthchecks/commit/166115ebfbcffd68ce7f6d2494657fff0e70575d


cuu508:
Ah, forgot about the API docs!
It would be slightly more precise to say "30 days" instead of "1 month".

brianstorti:
Yeah, that's a good point. Just updated it.

cuu508:
Nice one, thanks!

PR #165, opened 2018-04-29
Setting api key will save it as a byte

MounirMesselmeni:
urlsafe_b64encode return a byte, which will result in an api key saved with something like b'apikey'
I did not tested it but [urlsafe_b64decode](https://docs.python.org/3/library/base64.html#base64.urlsafe_b64decode) looks like doing the job also.


cuu508:
Good catch, thanks!

PR #166, opened 2018-04-30
Use mark_safe to use html in users list admin

MounirMesselmeni:


hc/accounts/admin.py
```diff
@@ -119,6 +120,7 @@ def engagement(self, user):
 
     engagement.allow_tags = True
```
cuu508: With the `@mark_safe` annotation, looks like this and `checks.allow_tags = True` is not needed any more.

MounirMesselmeni: You're right, it looks like Allow tags support has been removed in Django 2.0

PR #178, opened 2018-06-28
Fix for Python 3.6

danielfaulknor:
smtpd doesn't work with python3.6

Note, I haven't tested this with python 3.5 yet. Let me know if you're interested in merging and I will test.


cuu508:
I'd be happy to merge it.
I'm using both 3.5 and 3.6 in production, but am not using `manage.py smtpd`, so didn't notice the issue on 3.6

danielfaulknor:
Why I thought switching my healthchecks docker containers from bionic to xenial at 3am was a good idea, I'll never know... but I can confirm my changes still work on Python 3.5

PR #179, opened 2018-06-29
Improved the gitignore file

mwanjajoel:
Imrpoved the gitignore file to cater for other python dependencies that need to be ignored

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,110 @@
+
+# Byte-compiled / optimized / DLL files
 __pycache__/
-*.pyc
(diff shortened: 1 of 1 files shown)


cuu508:
Are you submitting this PR to the correct repo?
If you are, why do you need these .gitignore entries, what's your setup & workflow like?

mwanjajoel:
Hey @cuu508, yes I am. I added that .gitignore file since this is a python project and also since I find that its progressive and this might be forgotten 😎

PR #199, opened 2018-11-07
Allow specific channel uuid to be specified in create/update check api

muff1nman:


muff1nman:
Related - I would also like to see a channels api. At least being able to list the channels would be an important piece so that one can find the uuid to use in check update call as proposed in this PR. Does it make sense to return uuid in a channel list api response? Or should a name field be added instead and this PR changed to use channel name?

hc/api/views.py
```diff
@@ -81,8 +82,14 @@ def _update(check, spec):
     if "channels" in spec:
         if spec["channels"] == "*":
             check.assign_all_channels()
-        elif spec["channels"] == "":
+        else:
             check.channel_set.clear()
+            if spec["channels"] is not None and spec["channels"] != "":
+                channels = []
+                for raw_channel in spec["channels"].split(","):
+                    channel = get_object_or_404(Channel, code=UUID(raw_channel))
```
cuu508: If any of the channel codes is invalid, would be good to return 400 instead of 404.

cuu508:
Thanks for the PR! I'll review properly as soon as I can.

I think referencing channels by their `Channel.code` field (like you have done) would be the best.
Channels don't have good unique names at the moment. Referencing them as "slack" or "webhook" would not work, as an account can have multiple Slack, webhook etc. channels.

hc/api/views.py
muff1nman: Sure. Changed to respond with 400.

PR #211, opened 2019-01-07
Upgrading compatibility of Django with MySQL in python3.

Light-Gao:
1. Adding essential dependencies in `requirements.txt`, including `pymysql`;
2. Adding code in `hc/__init__.py` to avoid Django looking for MySQLdb automatically;
3. Adding some tips when people do not use default storage(sqlite);


cuu508:
I'm not too familiar with MySQL side of things when using Django. For running Travis tests we currently use `mysqlclient` (see `.travis.yml`). What are the pros and cons of using pymysql instead of mysqlclient?

requirements.txt
```diff
@@ -1,6 +1,16 @@
...
+idna==2.7
 psycopg2==2.7.5
+PyMySQL==0.9.3
+python-dateutil==2.7.5
 pytz==2018.7
+rcssmin==1.0.6
 requests==2.20.1
+rjsmin==1.0.12
+six==1.12.0
+urllib3==1.24.1
```
cuu508: Is it a recommended practice to list sub-dependencies in the top-level `requirements.txt`?

hc/__init__.py
```diff
@@ -0,0 +1,2 @@
+import pymysql
+pymysql.install_as_MySQLdb()
```
cuu508: I would prefer to have this in `settings.py` and execute it conditionally if `os.getenv("DB") == "mysql"`

Light-Gao: Yes it would be more specific and accurate.

requirements.txt
Light-Gao: I think it's more convenient to maintain this `requirements.txt` when listing all dependencies.  
Since we could just execute `pip freeze > requirements.txt` to finish it after coding for a long time.

Light-Gao:
As I know, `mysqlclient` is a fork of `MySQLdb` since `MySQLdb` does not support python3.x. `pymysql` is more stable  that `mysqlclient` and it's widely used when python applications communicate with MySQL.

requirements.txt
cuu508: I'm not convinced – what specific workflows / use case become more convenient?

A quick google search turned up this: https://medium.com/@tomagee/pip-freeze-requirements-txt-considered-harmful-f0bce66cf895

PR #214, opened 2019-01-22
fix api docs to point to project settings instead of account settings

kumikoda:
For a while I thought I had to pay in order to get the API key. Turns out it was just in the project settings page instead. 

You should consider adding a row in the pricing page to indicate that api access is included in the free tier as well. 

Thanks for creating such a great product! Go HC!

### DIFF
--- a/templates/front/docs_api.html
+++ b/templates/front/docs_api.html
@@ -59,8 +59,8 @@
 API key. By default, an user account on {% site_name %} doesn't have
 an API key. You can create one in the
 {% if request.user.is_authenticated %}
-<a href="{% url 'hc-profile' %}">Account Settings</a>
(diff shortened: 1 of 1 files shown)


cuu508:
Thanks, a good catch!

The API keys used to be in Account Settings. I'm in the process of adding multiple project per account support, and each project would have its own API keys and its own team. So I had to move this out to a Project Settings page, and had not updated the docs yet.

I don't think `{% url 'hc-project-settings' %}` alone would work, that route requires an UUID argument. I'll change the text to "Project Settings" and remove the hyperlink for now. Once the multiple projects feature is ready, docs will another update mentioning that API keys are per-project, not per-account.

Good point about mentioning API access in the pricing table, thanks!

PR #217, opened 2019-02-01
Allow SMTP configuration via environment variables

timfreund:
This change allows SMTP configuration via email, but users can still choose to configure the settings directly in local_settings.py

This is helpful for docker-compose workflows:  we configure a local mailhog container for development and we use a mailgun account for production.  The account settings for each are passed in via deployment specific environment files.

I'm open to suggestions for improving this PR.  Thanks!


cuu508:
Looks good, thanks!

PR #228, opened 2019-03-12
Improve environment variables support

toliger:
Makes it possible to define the path of the sqlite file and adds the environment variables "USE_PAYMENTS" and "REGISTRATION_OPEN"

### DIFF
--- a/hc/settings.py
+++ b/hc/settings.py
@@ -31,9 +31,8 @@ SECRET_KEY = os.getenv("SECRET_KEY", "---")
 DEBUG = envbool("DEBUG", "True")
 ALLOWED_HOSTS = os.getenv("ALLOWED_HOSTS", "*").split(",")
 DEFAULT_FROM_EMAIL = os.getenv("DEFAULT_FROM_EMAIL", "healthchecks@example.org")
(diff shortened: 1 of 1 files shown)


cuu508:
For `DB_PATH`, maybe use it as-is, without appending "hc.sqlite". In case somebody wants to use a filename other than "hc.sqlite".

For the description, could borrow this from Django docs: "Full path to the sqlite database file. When specifying the path, always use forward slashes, even on Windows."

toliger:
Thank you,

I took in your peace of advice, I renamed `DB_PATH` to `DATA_PATH`.

I changed the description in order to create a more generic folder.

cuu508:
I'd prefer to use something like

    'NAME': os.getenv("DB_NAME", BASE_DIR + "/hc.sqlite")

The DB_NAME variable would need to contain the absolute path including the filename. If DB_NAME is not set then it would default to "hc.sqlite" in project's directory.

I would name it DB_NAME to follow the existing naming convention: DB_HOST, DB_PORT, DB_USER etc.

toliger:
Thank you, it's done.

cuu508:
Thanks!

PR #229, opened 2019-03-14
Add Pager Team integration

Furchin:
This PR adds integration support for [Pager Team](https://pagerteam.com). ~I'm not convinced I got the database migration part right though!~ Database migrations now updated. Here are screenshots from it running locally:

Screen Shot 2019-03-13 at 10 09 42 PM
Screen Shot 2019-03-13 at 10 15 06 PM
Screen Shot 2019-03-13 at 10 16 12 PM


Furchin:
D'oh! Let me fix that coverage.

Furchin:
_sigh_ I think https://github.com/healthchecks/healthchecks/pull/229/files#diff-2a1f9332e08c8f7d00dfd227d0c6fa7fR297 should be covering the lines https://coveralls.io/builds/22187301/source?filename=hc%2Ffront%2Fviews.py#L754 coveralls claims aren't executed. I suspect coveralls is correct, but I'm not sure why they're not getting executed. If anyone wants to have a peek and see if I missed something I'd greatly appreciate it!

Furchin:
Whoops, forgot to `git add` the test file!

cuu508:
Hey @Furchin, awesome, thank you for contributing this! Just tried out the new integration and have a few comments.

The main one is to do with creating and resolving incidents. From a brief look, it looks like the Pager Team webhook creates an incident (in case there isn't an active one with the exact same name), but it doesn't support resolving the incidents.

Meanwhile the Healthchecks integration will call the webhook both when a check *goes down* and also when it goes *back up". So if I have a check named "Staging Backup", I might first see an incident "Staging Backup is DOWN" and then some time later a new, separate incident "Staging Backup is UP". 

I think it would make sense to only call the webhook when checks go down. That can be done easily in the `hc.api.transports.PagerTeam` class, take a look at the `hc.api.transports.Sms` for an example.

The other comments I have are minor and cosmetic:

 * The UI says "Pager Team" but also "PagerTeam" in a few places: `templates/front/channels.html` and `hc/api/models.py`
* The logo and the `setup_pagerteam_1.png` both look a bit grainy. I can fix both, but for the instructions screenshot could you send me a full sized version?
* In the landing page, Pager Team has the subtitle "On-call rotations without limits". I'd like the subtitles to show the service "category": chat, push notifications, incident management. Would you be OK with changing the tagline to "Incident Management"?

Furchin:
Ok I added high-res versions of both PNGs (I had resized them to be comparable to the existing logo and instruction images, but yeah they were grainy). Feel free to re-size or crop as you see fit. Updated "PagerTeam" to "Pager Team" consistently, and updated the subtitle as well following the other on-call management tool subtitles, but if you prefer "Incident Management" that's fine too. 

As for re-triggering notifications, I made a change to Pager Team to ignore resolve event_types. I think it's better to receive and ignore them than to not receive them at all -- affords the opportunity to implement auto-resolve in the future (though I think philosophically I disagree with doing so but that's a topic for a longer debate).

cuu508:
Thanks for making these changes! 

> As for re-triggering notifications, I made a change to Pager Team to ignore resolve event_types. 

Perfect, that's even better.

A heads up, I'm on vacation so might take me some time to review/merge/deploy these changes.

Furchin:
Totally fine, have a lovely vacation! :)

PR #241, opened 2019-04-15
Update README.md

evaryont:
Since anonymous checks aren't created any more, don't document the manage command to clean them up.


cuu508:
Thanks!

Issue #271, opened 2019-08-06
Support Apprise to Simplify all of your Notification Requests


caronc:
I recently had an [enhancement request opened for Apprise](https://github.com/caronc/apprise/issues/139) to support your service.  But having visited your site and seeing you have already open requests for:
* Gotify #270 
* Twilio #222 
* Zulip #202 
* Flock #159 

It might make more sense to just add an `APPRISE_URL` variable inside your `hc/settings.py` file and kill all your birds with one stone (and more).  It should be relatively trivial to do since [Apprise](https://github.com/caronc/apprise) is python based as well. 

Thoughts?

cuu508:
Thanks for the suggestion and the PR – it looks simple (good!) and clean.

Apprise looks like a well-designed, versatile tool for sending notifications, but I have a few concerns about integrating it:

* **Credentials in Apprise URLs.** I think what's likely to happen is people will put things like their root AWS keys in the URLs. Which increases the liability & risks for me. Obviously the existing integrations already contain credentials and secrets, but they usually have a tighter scope, and fewer are needed. For example, user cannot and doesn't need to specify AWS keys for sending emails.

* **Security risks for the hosting platform.** People are free to enter any Apprise URLs they like. At a quick glance through the supported notification services, they could send DBUS messages on the local server. Or post JSON payloads to `localhost:anyport`. I would need to make an extensive audit of all supported notification services (or white-list them one by one) to be comfortable running this.

* **Extra dependencies.** Apprise brings in PyYAML, makrdown and oauthlib. Not a big issue, but worth pointing out.

* **Duplication of functionality.** Two ways to send a Slack notification, two ways to send a Telegram message etc. On one hand, it's more power to the user, on the other hand it's more decisions (which method to use?) and potential for confusion. 

* **User friendliness.** The Apprise notifications would be great for expert users, but could be problematic with non-technical users ("is it a service or library?", "It asks me to enter an URL with a token. Where do I get the token?"). 

One idea I have is – how about making this integration optional? An boolean `APPRISE_ENABLED` flag in `settings.py` or similar. This way, operators of self-hosted Healthchecks instances could make their own decision whether to use it or not.

caronc:
Your concerns are valid. If I can just comment on a few of them though:
* **As per some of your security points**:
[...]

PR #272, opened 2019-08-07
Apprise Integration

caronc:
Added Apprise Integration into Healthchecks which references #271 
Screenshot from 2019-08-07 19-11-58

Also additionally fixed a minor typo _PagerTeam_ http post (_fixed Content-Type header_).


immanuelfodor:
This is huge, Apprise opens up a lot of other integration options, thank you for the work you put into this.

PR #291, opened 2019-10-01
Allow setting default timezone via env variable

SuperSandro2000:


cuu508:
Before accepting this PR, I'm interested in what specific behavior of the app do you want to change. I'm assuming the app is displaying UTC timestamp somewhere in the UI, and you want it changed to a different timezone. What places / what situations are you thinking of specifically?

For reference, from [Django docs](https://docs.djangoproject.com/en/2.2/ref/settings/#std:setting-TIME_ZONE):

>  TIME_ZONE. [...] When USE_TZ is True, this is the default time zone that Django will use to display datetimes in templates and to interpret datetimes entered in forms.

SuperSandro2000:
> What places / what situations are you thinking of specifically?

I wanted to change the default `Server's Time Zone` when creating a new check.
All my services use the same timezone and I found it annoying to change it all the time.

cuu508:
This would have to be implemented differently. 

Setting TIME_ZONE in settings would change the default timezone used for display in a few places, e.g. Django admin. But it would have no effect on the default, pre-selected timezone when configuring a check.

PR #293, opened 2019-10-03
feature: golang code snippet added

omurbekjk:


cuu508:
Thanks for the PR! 

I was looking into whether `resp.Body.Close()` is optional or not, and found this SO answer: https://stackoverflow.com/a/18601625/5821

It says you can use the HEAD head method and then there's no worries about handling the response. Healthchecks *does* handle HEAD requests, so I think this nippet (and probably the snippets for other languages too down the line) could be changed to use HTTP HEAD.

omurbekjk:
Yeah, agree. @cuu508 Should I change it to head instead?

cuu508:
@omurbekjk yes, please do!

omurbekjk:
I updated

PR #313, opened 2019-12-17
set/update the checks description through the API

### LINKED ISSUE #311: API ability to set the check description
I'd like to update the checks' description from the API.
Could that be implemented?

brammeleman:
closes #311


cuu508:
Thanks, looks good! 

I'd also want to add `desc` in `hc/api/schemas.py` (for type validation) but that's a quick fix, I'll do that myself.

brammeleman:
> Thanks, looks good!
> 
> I'd also want to add `desc` in `hc/api/schemas.py` (for type validation) but that's a quick fix, I'll do that myself.

Thanks!!
/Bram

PR #320, opened 2020-01-16
Increase allowable length of Matrix room alias

jerrykan:
The existing 40 character limit prevents using the integration will Matrix servers that might have a fairly lengthy hostname (ie. 'matrix.internal.example.com' would only allow 12 characters for the
room name or ID, and room IDs are 19 characters long).

Increasing the `max_length` to `100` is still fairly arbitrary but itdoes match up with the `max_length` of the `name` field of the `Channel` model and should cover most instances.


cuu508:
Looks good, thanks for the PR!

PR #325, opened 2020-01-30
Updated default port in readme

samyerkes:
Updated default port in README.md to 8000


SuperSandro2000:
👍 

See https://github.com/healthchecks/healthchecks/blob/master/hc/settings.py#L151

cuu508:
Good catch, thanks!

PR #342, opened 2020-03-14
Introduce Project Name in Slack notification

sairam:


cuu508:
@sairam how about naming the field just "Project", not "Project Name"?

PR #358, opened 2020-04-08
Using existing function getAmount

lobovkin:


cuu508:
That's a good catch, thank you!

I didn't expect to receive PRs for billing code! Do you plan to use the billing code somewhere, or did you just spot this by reviewing recent commits?

lobovkin:
Hello. I plan to add cron job monitoring functionality to existing website monitoring project. So I checked the billing code to understand how to integrate. Now I see that it is easier to modify limits directly in the database.

PR #360, opened 2020-04-18
Remove redundant '-X POST' to curl

bdd:
Passing `--data-raw` to curl implies the request is method will be POST.
Unless we intend to do something entirely different, -X method override
shouldn't be used.

Curl's author Daniel Stenberg (@bagder) wrote about this back in 2015
https://daniel.haxx.se/blog/2015/09/11/unnecessary-use-of-curl-x/


bdd:
If uses the same source repo, healthchecks.io site will require a deploy too.

cuu508:
TIL – thank you!

PR #366, opened 2020-05-01
Add runitor to resources.{md,html}

bdd:
From its README:

Why Do I Need This Instead of Calling curl from a Shell Script?

In addition to clean separation of concerns from the thing that needs to
run and the act of calling an external monitor, runitor packs a few neat
extra features that are bit more involved than single line additions to
a script.

It can capture the stdout and stderr of the command to send it along
with execution reports, a.k.a. pings. When you respond to an alert you
can quickly start investigating the issue with the relevant context
already available.

It can be used as a long running process acting as a task scheduler,
executing the command at specified intervals. This feature comes in
handy when you don't readily have access to a job scheduler like crond
or systemd.timer. Works well in one process per container environments.


bdd:
Direct link to repository: https://github.com/bdd/runitor

cuu508:
Wow, runitor is a great idea! If you want to use all features (start signal, retry parameters, capture and submit script output) with curl, it starts to look pretty busy. runitor nicely hides that complexity away from the shell scripts and cron job definitions.

I gave it a quick test, and am looking forward to testing it some more.

PR #380, opened 2020-06-11
Switched from pipeing to --output /dev/null for curl

Simonmicro:
- That way the user could still receive logs e.g. by using the traditional `MAILTO` attribute of cron.
- Maybe the system would deliver the outputs over the default notification-system - the usage of the pipe operator breaks this for every command in the chain.


cuu508:
Makes sense to silence only curl's output – thanks!

Issue #349, opened 2020-03-26
API: Return history for a single check


jameskirsop:
Would be great to allow a `?history=x` query variable, where `x` is the number of days you wish to retrieve historical data for to the single end point, so it's possible to show availability over time.

Will try and write this up in a new branch and do a PR.

jameskirsop:
Realised that this is going to be less useful than I originally thought because there's no events stored for failures, unless there's alerting set up.

@cuu508, have you considered storing a failure event for checks that pass their grace time without a ping?

cuu508:
To clarify, with the `?history=x` parameter, would you want the API return JSON with the same structure, but with values from x days ago? 

We're not tracking the edit history of things like check's name, tags, schedule etc., so this would be difficult to implement properly.

jameskirsop:
No, I'm not wanting to track edit history, but status history.

To illustrate, at the moment, if you look at the status page for a single check (`/checks/<uuid>/details`) you are presented with a Log view. It's the data of this log view I would like to return via an additional 'history' attribute on the returned JSON object.

However, if you have no notifications setup, that Log view shows all the successful Pings, but none of the times where it's passed the grace period (i.e. failed according to the configured parameters for that check).

I'm suggesting that a new type of event be added to the _Log_ that show a non-ping failure event, not just when a notification is sent, but any time a failure occurs. Perhaps this _could_ be done by creating a new `Ping` object with the `fail` status when the grace period passes?

The `history` attribute on the returned JSON object (which would just be an addition to the existing attributes) would contain an ordered list of events with timestamps and event `kinds`, similar to the summary in the `details` view.

jameskirsop:
I've [built something that fulfils](https://github.com/jameskirsop/healthchecks/tree/return-single-history) my needs by using a query on the `Flips` objects attached to a check.

These changes return a JSON object that looks like this:
```
{
    "name": "My First Check (copy)",
    "tags": "",
    "desc": "",
    "grace": 3600,
    "n_pings": 1,
    "status": "down",
    "last_ping": "2020-03-23T23:30:18+00:00",
    "next_ping": null,
    "history": [
        {
            "timestamp": "2020-03-23T23:30:18.767Z",
            "status": 1
        },
        {
            "timestamp": "2020-03-25T00:30:18.767Z",
            "status": 0
        }
    ],
    "unique_key": "e855898bebff1756cde7c571319d877d07a38dab",
    "timeout": 86400
}
```

[...]

PR #381, opened 2020-06-12
Return single history

jameskirsop:
This PR contains changes to address #349 in light of discussion there.

### DIFF
--- a/hc/api/views.py
+++ b/hc/api/views.py
@@ -1,4 +1,5 @@
 from datetime import timedelta as td
+from datetime import datetime
 import time
 import uuid
 
@@ -8,6 +9,7 @@ from django.http import (
     HttpResponse,
     HttpResponseForbidden,
     HttpResponseNotFound,
+    HttpResponseBadRequest,
     JsonResponse,
 )
 from django.shortcuts import get_object_or_404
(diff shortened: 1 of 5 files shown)


hc/api/views.py
```diff
@@ -291,6 +295,61 @@ def pings(request, code):
...
+    #         history_start,history_end
+    #         )
+    #     ))
+
+@cors("GET")
+@csrf_exempt
+@validate_json()
+@authorize_read
+def get_flips_by_unique_key(request, unique_key):
+    checks = Check.objects.filter(project=request.project.id)
```
cuu508, line 348: This will fetch a Check instance first by the unique_key, and then the second time in hc.api.views.flips by code. 

We can avoid a redundant SQL query and redundant view decorators by adding another function which both views call. Views would be responsible for locating the check (either by code or by unique_key), and the third function would be responsible for the rest: parsing timestamps, fetching flips and returning them.

hc/api/views.py
```diff
@@ -291,6 +295,61 @@ def pings(request, code):
...
+    check = get_object_or_404(Check, code=code)
+    if check.project_id != request.project.id:
+        return HttpResponseForbidden()
+
+    if all(x not in request.GET for x in ('start','end','seconds')):
+        flips = Flip.objects.select_related("owner").filter(
+            owner=check, new_status__in=("down","up"),
+        ).order_by("created")
+    else:
+        if any(x in request.GET for x in ('start','end')) and 'seconds' in request.GET:
```
cuu508, line 312: Can we rewrite this (and the previous `if`) in a simpler and a dumber way?
IIUC this says "if (either start or end is present) *and* seconds is present that's an error". A simpler way to express that would be:

    if "seconds" in request.GET and ("start" in request.GET or "end" in request.GET):

hc/api/views.py
```diff
@@ -291,6 +295,61 @@ def pings(request, code):
...
+@csrf_exempt
+@validate_json()
+@authorize_read
+def flips(request, code):
+    check = get_object_or_404(Check, code=code)
+    if check.project_id != request.project.id:
+        return HttpResponseForbidden()
+
+    if all(x not in request.GET for x in ('start','end','seconds')):
+        flips = Flip.objects.select_related("owner").filter(
```
cuu508, line 308: Unless I'm missing something, don't need to `select_related("owner")`

hc/api/views.py
```diff
@@ -291,6 +295,61 @@ def pings(request, code):
...
+
+        if 'start' in request.GET:
+            history_start = datetime.fromtimestamp(int(request.GET['start']))
+        if 'end' in request.GET:
+            history_end = datetime.fromtimestamp(int(request.GET['end']))
+
+        if 'seconds' in request.GET:
+            history_start = datetime.now()-td(seconds=int(request.GET['seconds']))
+        
+        flips = Flip.objects.select_related("owner").filter(
```
cuu508, line 326: A perhaps cleaner alternative:

```
flips = Flip.objects.filter(owner=check)

if "start" in request.GET:
    flips = flips.filter(created_gt=...)

if "end" in request.GET:
    flips = flips.filter(created_lt=...)

if "seconds" in request.GET:
    flips = flips.filter(created_gt=...)
```

hc/api/views.py
```diff
@@ -189,6 +191,8 @@ def get_check(request, code):
     if check.project_id != request.project.id:
         return HttpResponseForbidden()
 
+    if 'history' in request.GET:
```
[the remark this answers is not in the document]
jameskirsop: Now obsolete, removing! :) Good pick up.

hc/api/views.py
```diff
@@ -291,6 +295,61 @@ def pings(request, code):
...
+        
+        flips = Flip.objects.select_related("owner").filter(
+            owner=check, new_status__in=("down","up"),
+            created__gt=history_start,
+            created__lt=history_end
+        ).order_by("created")
+
+    dictStatus = {"up":1,"down":0}
+
+    return JsonResponse({"flips": list(map(lambda x: {'timestamp':x.created,'up':dictStatus[x.new_status]}, flips))})
```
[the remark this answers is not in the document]
jameskirsop: Good suggestion, applying changes.

Issue #404, opened 2020-07-17
full localization support


peppelinux:
Hi guys, I'd going to translate healthchecks in italian.
For doing it I'd have to put {% trans %} template tags in all the .html files.
Are you interested in it? I'd push a pull request with this and also a italian localization
regards

cuu508:
Just to understand your use case, are you planning to use Healthchecks in a setting where there's a requirement that the software is localized? If you go through with the translation, would it be an one time thing, or would you be planning to maintain it (translate new and updated strings, test translations) long term?

There's a bunch of pros and cons and honestly I'm not sure what's the correct approach. 

Setting the project up for translations is  a fair bit of work. It is not only {% trans %} tags, there are strings also in Python and JS code. If we do it, I'd like to do the changes incrementally, not one big PR that changes pretty much all files in the project.

Having the {% trans %} tags, and gettext stuff in Python code adds some performance overhead, but more importantly it adds complexity and verbosity. It may may or may not be worth it, depending on how many people are going to use the non-English versions.

One project I sometimes look at to see "how others are doing it" is Sentry. Sentry does support translations, in "User Settings" the user can select their UI language. I tested German and Spanish, and in both cases maybe 20% of UI text changed, it was still mostly English. I would guess that the translations were "complete" at some point, but as the project evolved the translations didn't keep up. Sentry is *much* bigger project than Healthchecks, so my worry is that same thing would happen here.

peppelinux:
Thank you @cuu508 for your reply,
I often use django in my projects and I'm quite experienced with localization support.
In my organization we really appreciate healthchecks and I want first of all spend this moment to thank you all for the wonderful work done, sincere congratulations.

We made a fork and we just localizaed menu voices (base.html) and welcome page, because this was needed in my organization. If we would create a branch for a full localization I can do the italian part, english will be the default one and don't need any translations, other languages would be merged as third contribution come, if come.

Probably only the welcome page and the menu voices could be localizaed, see our page here:
https://healthchecks.unical.it 
set italian language in your browser (or install an addon to switch languages)

our repo is here:
https://github.com/UniversitaDellaCalabria/healthchecks

I'd merge your latest commits and update the locale branch and then to a PR from this to the branch you will indicate to me, if interested. I also added SAML2 Support with djangosaml2 if you are interested in a native support of this in healthchecks.

Thank you again and continue do this way, it's awesome

cuu508:
[...]

PR #405, opened 2020-07-21
Localization support

peppelinux:
This is the first step for a localization aware support in healthchecks
https://github.com/healthchecks/healthchecks/issues/404#issuecomment-661750282

- welcome and page html page with block trans
- italian localization


peppelinux:
This was a transitional PR that would be improved with a following one

Issue #277, opened 2019-08-21
Support multiple language


roperluo32:
Do we have any plan to support multiple language？

cuu508:
So far there haven't been any requests for localization support, so I have no near-term plans for that. 

I've assumed the audience of this tool is technical and so is generally fine with English. What languages are you interested in? Are you dealing with a regulatory requirement?

roperluo32:
Thank you for your replay.
I want to translate this web to Chinese. I found this web don't support multiple language template now. So may be we can make the django html support language template.

cuu508:
In #406, we set up the landing page  for translation, and added translations in Italian and Latvian. You can submit a PR for translations in other languages. 

I'm happy to review PRs that add translation support for other pages as well.

PR #406, opened 2020-07-22
IT localization

peppelinux:
As discussed here:
https://github.com/healthchecks/healthchecks/issues/404

Thsi would be also a first step for:
https://github.com/healthchecks/healthchecks/issues/277


cuu508:
Looks good, thanks!

Do you plan to translate other pages / screens too?

peppelinux:
Concretely not at this moment, but It would be trivial when we'll need It. Thank you for all

peppelinux:
Ah, Just a Little issue. Can't render {{ site_name }}. Is there some template contest processore to be loaded?

cuu508:
Yes, `site_name` should come from this template context processor: https://github.com/healthchecks/healthchecks/blob/master/hc/settings.py#L91

peppelinux:
> Yes, `site_name` should come from this template context processor: https://github.com/healthchecks/healthchecks/blob/master/hc/settings.py#L91

Very thankfull to you @cuu508, I missed to add that file in my last remote merge. Regards

PR #426, opened 2020-09-10
Modify format of service name in Teams notification

nigelm:
Teams notifications have all text parsed as markdown.  If the service name has underscores in them it causes the service name to be misrendered.

This change puts backquotes around the service name, which formats it as fixed width on a grey background, with underscores handled correctly and pulls your focus to the service name.

An example of what happens currently is attached.

There is a newer notifications message format that is recommended - see https://docs.microsoft.com/en-us/outlook/actionable-messages/message-card-reference - but I'll experiment with that offline

### DIFF
--- a/templates/integrations/msteams_message.json
+++ b/templates/integrations/msteams_message.json
@@ -3,7 +3,7 @@
     "@type": "MessageCard",
     "@context": "https://schema.org/extensions",
     "themeColor": "{% if check.status == "up" %}5cb85c{% endif %}{% if check.status == "down" %}d9534f{% endif %}",
-    "text": "“{{ check.name_then_code|escapejs }}” is {{ check.status|upper }}.",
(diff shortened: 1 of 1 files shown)


nigelm:
Forgot the example screenshot...

Screenshot 2020-09-10 at 14 42 56

This change would make the second line be rendered as:-

`ntp_gps_unit_ping` is DOWN

cuu508:
Thanks for reporting the issue and the PR!

Wrapping the name in backticks is problematic if the name itself contains a backtick.

But, to be fair, "foo_bar_baz" is a much more likely name than "foo\`bar\`baz".

I just pushed a fix that escapes more characters that have special meaning in Markdown. The test string I used was:

```
TEST _underscore_ `backticks` <u>underline</u> \backslash\ "quoted"
```

And it now appears in Teams like so:

image

There may be edge cases that I have not found yet. 

Annoyingly, looks like the Slack integration has a similar issue, but handles Markdown slightly differently. So more work there.

Issue #467, opened 2021-01-11
Move self-hosting instructions from README to online docs


Romanmir:
I'm not seeing where to configure options for Pushover notifications.

I already have a notification set up that pings me on backup failure, but I'd like to set up a new notification for a separate project that notifies me in an emergency fashion when an important process is missed.

Can you expand on the notes that you have in the "Configuring Notifications" page?

Here are the questions that should be expanded; "Where do I configure the push settings for pushover?", "Are the settings for pushover per project?", "Can multiple settings 'profiles' be used within a specific project?"

Thanks for making this software, I use it consistently.

cuu508:
Thanks for the feedback!

> Where do I configure the push settings for pushover?

You can set the push settings (priority) when adding the Pushover integration. There is currently no option to edit the integration settings, after the integration is added.

You can, however, assign checks to integrations (tell which checks should use which integrations) at any time from 3 places:

* from the list of checks, by clicking on the integration icons
* from the check details page, by clicking on the ON / OFF buttons in the "Notification Methods" section
* from the integrations list, by clicking on cells in the "Assigned Checks" column

> Are the settings for pushover per project?"

They are per-integration (integration = "settings profile"). Each integration belongs to a specific project. If you have multiple projects, and want to use Pushover in each, you must add a Pushover integration to each project.

> Can multiple settings 'profiles' be used within a specific project?

Yes, you can add multiple integrations (settings profiles) of the same type to a single project. For example, you can have three Pushover integrations for low, medium and high priority within a single project. Or you can have multiple Pushover notifications that notify different people.

--

Some of the above is already in the docs, some isn't. I'll review the docs and see what can be improved.

Romanmir:
Ok, but _how_ do I change the settings when setting up the integration?

Maybe I'm missing something. On my integrations list, I do not see Pushover. I've been using the Site Administration interface to do this. I'm using the self-hosted docker image. What am I missing?

cuu508:
Ah, sorry, I didn't realize you don't see the integration in the list at all!

README has instructions for setting up the Pushover integration: https://github.com/healthchecks/healthchecks#pushover

Romanmir:
Ok, I think I'm on the right track now. I think my feedback would change to be: I feel like the contents of the README.md file should be fully integrated into the online docs.

cuu508:
The primary audience of the online docs is the end users of the service. I think publishing README as part of the online docs would just cause confusion. 

[...]

PR #468, opened 2021-01-17
Added some of the README to the online docs

Romanmir:
This is to address #467

Integrated the "Integration" section from the READEME.md into the configuration_notifications.html file. I was less worried about the configuring_notification.md file, but I can update that one as well if needed. I'm using the "code" html tag to highlight some of the technical parts in the html file. I didn't see that you were using that tag in any of your css files.

As I'm using the docker version, I wasn't able to preview this in a live environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,7 +374,7 @@ https://api.slack.com/apps/APP_ID/oauth?).
 
 To enable Discord integration, you will need to:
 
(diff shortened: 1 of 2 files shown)


cuu508:
The primary audience of the online docs is the end users of the service. I think publishing README as part of the online docs would just cause confusion.

PR #470, opened 2021-01-23
Adding French Locale

msansen1:
Hi,

I would like to try healthcheck and have seen that french wasn't yet supported. So if you want you can add french to your awesome solution.

Regards.

Mat


cuu508:
Looks good  – thank you! 
Merged it via the command line.

PR #471, opened 2021-01-23
Feature/integrations toggle

msansen1:
Hi,

In my use case, i need to be able to disable some integrations to my users so i propose you some changes enabling to toggle integrations to disabled it inside settings.py

with this, Healthcheck admin is able to switch some integration Off with env like:
DISCORD_ENABLED=False;SLACK_ENABLED=False

If nothing is pass to env, nothing change

it hide Discord and Slack from available integrations.

Hope you will agree with it.

Regards

Mat


cuu508:
Thanks for the PR!

I like the general idea, and think the ability to disable certain integrations would be useful.

Before I can merge, I'd like to make a few changes. 

* Not all integrations need an explicit _ENABLED setting. For example, Pushover: we consider Pushover as enabled if and only if `PUSHOVER_API_TOKEN` is set. And so, I think, `PUSHOVER_ENABLED` is redundant. Same for Pushbullet, Matrix, Discord, Telegram, SMS, Whatsapp, Trello, and Line Notify.
* This PR changes the default value of SECRET_KEY. Can you revert that change? I want the SECRET_KEY be blank to reduce the chance of somebody running this in production with the default, not-secret, SECRET_KEY value.
* The PR includes French translations, (thanks for that!), but they are already in a separate PR. Can you please remove them from this PR?

msansen1:
Hello,

Thanks for the Review, i just empty the secret key (maybe add an info in the readme that said it must not be empty because project can't start until it's empty)
I remove French since you already accept it (Think some files are missing because i only give you the translations without "compiled .mo" version) need compilemessages to work.
About explicit _ENABLED, i don't have already revert: Don't you think it can be more clear for administrator to use this same setting _ENABLED for all the integrations?
in facts changes have to be done in hc/front/views.py to only take ENABLED AS feature enabler.

I leave you the choice and will change as you prefer. I just think that it can be more "universal"

Regards.

Mat

cuu508:
> I leave you the choice and will change as you prefer. I just think that it can be more "universal"

I see your point. On the other hand, you could have ambiguous states where `PUSHOVER_ENABLED=True` but `PUSHOVER_API_TOKEN` does not have a value, for example. I'd still like to go with the minimal changes, and add _ENABLED flags only for the integrations that don't have any other configuration parameters.

msansen1:
i revert my changes on existing settings and fix tests on existing settings.

Mat

cuu508:
Hi Mat,

This still has a few problems – a failing test because of a pushbullet / pagerduty mixup. The template conditionals for Slack and Pagerduty are not quite right. I'd also like to add individual test cases for each _ENABLED flag. 

I'll close this PR and make the changes in smaller chunks, using this as a blueprint.

msansen-esncf:
Hi

Ok, Sorry for the mixup, i don't know all theses tools and somes names are quite same making me errors.
  thanks for your help.

Mat

PR #483, opened 2021-02-23
Changing static height and width to be constraints

jameskirsop:
This commit allows for logo.png (and it's scaled alternates) to keep their own proportions while still being constrained in the existing bounds of the base layout for each page. Without this change, custom logos that don't exactly match the proportions (50x59, 50x200) are skewed to fill the bounds.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -77,16 +77,16 @@
             {% elif request.user.is_authenticated%}
                 <img
                     id="logo"
-                    height="50"
-                    width="59"
+                    max-height="50"
+                    max-width="59"
                     src="{% static 'img/logo.png'%}"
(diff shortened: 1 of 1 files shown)


cuu508:
I pushed a change that removes the width attribute. I think it accomplishes the same goal: if the logo isn't exactly 50x59 or 50x200, it allows the browser to preserve the aspect ratio. But it keeps the height fixed to 50px, as that's what the layout is designed for.

The reason for setting explicit width and height in HTML was to avoid a page reflow on the first page load when the logo isn't in browser's cache. But I think the height alone should be enough: since the height is still fixed, loading the logo should not cause the content to shift down vertically.

PR #485, opened 2021-02-28
fix line height on request body

cocide:
The line height in the ping request body is being set by the bootstrap.css body tag to 1.42857143. The additional line height makes request bodies with unicode characters have spacing between lines that shouldn't be there.
Before:
Before Commit
After:
After Commit


cuu508:
Yep, looks better and fits more information with your patch. Thanks!

PR #487, opened 2021-03-04
Add colored emojis to telegram messages

lhaus:
so you can see whether it's up or down at first glance


cuu508:
Great idea, thanks!

PR #490, opened 2021-03-09
Handling database connectivity issues

jameskirsop:
This PR is partially to scratch my own itch, but will be useful for the community.

It adds a bit of handling of database slowness to the `sendalerts` management command so that it the script doesn't immediately crash if your database of choice is unavailable, or in my case, can't secure a write lock in a timely fashion.

I've had the service for `sendalerts` crash recently due to some performance issues we've been experiencing with mariadb. While it's trivial to restart the service, we sometimes miss notifications during the time it's down and the time we kick the service back up again. The PR increases the robustness of the script to decrease the chance of this particular exception (immediately) causing a crash.

### DIFF
--- a/hc/api/management/commands/sendalerts.py
+++ b/hc/api/management/commands/sendalerts.py
@@ -4,6 +4,7 @@ from threading import Thread
 
 from django.core.management.base import BaseCommand
 from django.utils import timezone
+from django.db.utils import OperationalError
 from hc.api.models import Check, Flip
 from statsd.defaults.env import statsd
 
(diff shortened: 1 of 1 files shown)


cuu508:
>  While it's trivial to restart the service, we sometimes miss notifications during the time it's down and the time we kick the service back up again. 

Without this patch, when `OperationalError` is thrown, the `sendalerts` process crashes, right?

There are more places in the script that can throw a exceptions (for example, `process_one_flip` also does database queries).  I think the right way to go about this is to make sure `sendalerts` gets automatically restarted when it crashes. For the hosted service, I run it as a systemd service, and systemd takes care of restarting it, waiting some time between restart attempts etc.

PR #506, opened 2021-05-03
Add more information to the prunepings command.

yonas:
Add a reminder that 100 of the most recent pings were kept.


cuu508:
"Kept 100 most recent pings." – this can be confusing in some cases:

* The ping limit can have other value than 100. For example, on Healthchecks.io, paid accounts have ping limit of 1000
* If a check has less than 100 ping log entries (say, 20), then saying "kept 100 most recent pings" is not quite correct either.

PR #511, opened 2021-05-08
Dark mode

andyloree:
For your consideration, the following are the minimal CSS files (all suffixed with ```-dark```), placed after the corresponding existing CSS files in the ```base.html```, with the ``` media="(prefers-color-scheme: dark)"``` attribute.  This will respect browse/OS settings for dark mode rendering.

Certainly possible to change this behavior a per-user saved preference or similar code-driven option as opposed to browser preference driven, but I thought this would present a minimal change, no-code, option. 

One draw back to this implementation is any [older browser](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme#browser_compatibility) that doesn't support theme preference will render dark mode by default.  A workaround to this would be to leverage per-page JS to assist correct loading on older browsers, which I would be happy to add.

Below are screen shots of the dark mode in use:

**01 - Welcome Screen**
01-welcome

**02 - Login/Signup**
02-login

**03 - Docs**
03-docs

**04 - Projects**
04-projects

**05 - Checks**
05-checks

**06 - Check Details**
06-check-details

**07 - Integrations**
07-integrations

**08 - Badges**
08-badges

**09 - Settings**
09-settings-dialog

I tested using Firefox and Chromium, and emulated mobile rendering.  It is possible that may have missed a few dark-on-dark rendering issues (tried my best to test all pages/areas). 

Thank you for your consideration, and for making such an excellently simple tool for healthcheck monitoring.

Andrew

### DIFF
--- a/static/css/badges-dark.css
+++ b/static/css/badges-dark.css
@@ -1,0 +1,4 @@
+.table.badge-preview code {
+    background-color: #666;
+    color: #000;
+}
--- a/static/css/base-dark.css
+++ b/static/css/base-dark.css
@@ -1,0 +1,36 @@
+#logo {
+    max-height: 50px;
+    margin-top: -16px;
+    background-color: #aaa;
+}
+
+#nav-sign-up span {
+    display: inline-block;
+    color: #ccc;
+    background-color: #7ab694;
+    font-weight: bold;
+    border: 1px solid #1ea65a;
+    border-radius: 5px;
+    padding: 8px 16px;
+    transition: all 0.2s ease;
+}
+
+.hc-dialog {
+    background: #FFF;
+    color: #000;
+    padding: 2em;
+    margin-top: 100px;
+}
+
+.hc-dialog h1 {
+    margin: 0;
+    color: #000;
+}
+
+
+#debug-warning {
+    background: #333;
+    text-align: center;
+    font-size: small;
+    padding: 2px 0;
+}
--- a/static/css/bootstrap-dark.css
+++ b/static/css/bootstrap-dark.css
@@ -1,0 +1,382 @@
+/*!
+ * Bootstrap v3.3.6 (http://getbootstrap.com)
+ * Copyright 2011-2015 Twitter, Inc.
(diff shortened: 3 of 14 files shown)


cuu508:
Hi Andy, 

Thanks for taking the initiative and doing the work! I support the idea of adding a dark mode. Dark mode would be one more thing to think about and test when working on UI but on balance I think it would be worth it. I also think starting with the standard `prefers-color-scheme: dark` is fine, and we can consider adding JS-powered UI toggles later, maybe.

Putting the dark mode stuff in `-dark.css` files keeps stuff nice and organized, but I've been experimenting with a slightly different approach using CSS variables (custom properties):

1. define common UI colors as CSS variables in [variables.css](https://github.com/healthchecks/healthchecks/blob/dark/static/css/variables.css). For consistency, use the same naming that Bootstrap's variables.less uses. 
2. in variables.css, also define UI colors for the `prefers-color-scheme: dark` media selector
3. use the CSS variables throughout the rest of CSS files, instead of "hardccoding" the color hex values. [Example](https://github.com/healthchecks/healthchecks/blob/dark/static/css/projects.css#L2)
4. there's a problem that many colors are defined in bootstrap.css, and it would be problematic [1] to use CSS variables in bootstrap sources. My solution so far is to not change bootstrap.css, but instead add overrides on top of it that redefine colors of the standard bootstrap components. [See here](https://github.com/healthchecks/healthchecks/blob/dark/static/css/base.css#L223).

I just pushed a branch "dark" with the above changes. It's very much work in progress, and for now has various issues. For example, most hover states still use light colors. But some of the screens are already starting to look ok.

Projects:

image

Check details:

image

Settings:

image

What are your thoughts on this approach? Would you be interested in collaborating on it, and transfer some of your tweaks to it?


[1] bootstrap.css is generated from .less files. Base colors are defined in [variables.less](https://github.com/healthchecks/healthchecks/blob/dark/stuff/bootstrap/variables.less). We cannot replace these with `var(--some-color)` syntax because Bootstrap use less.js color functions (`darken`, `lighten` etc.) liberally and these need concrete color values to work.

PR #524, opened 2021-06-07
Close response body in Go example as recommended in http package documentation

hjr265:
The official Go documentation on the http package recommends that caller must close the body of the response.

https://golang.org/pkg/net/http/#Response

```
    // ...
    // The http Client and Transport guarantee that Body is always
    // non-nil, even on responses without a body or responses with
    // a zero-length body. It is the caller's responsibility to
    // close Body. The default HTTP client's Transport may not
    // reuse HTTP/1.x "keep-alive" TCP connections if the Body is
    // not read to completion and closed.
    // ...
    Body io.ReadCloser
```

Since the example already demonstrates the sane practice of not using the default HTTP client, it would make the example _complete_ if one more line is added to close the response body.

I suppose in a lot of use cases, this ping will happen from within a loop...? If that is the case, this would be even more relevant.

### DIFF
--- a/templates/docs/go.md
+++ b/templates/docs/go.md
@@ -14,10 +14,11 @@ func main() {
         Timeout: 10 * time.Second,
     }
 
-    _, err := client.Head("PING_URL")
+    resp, err := client.Head("PING_URL")
     if err != nil {
         fmt.Printf("%s", err)
     }
+    resp.Body.Close()
 }
 
 ```


cuu508:
Thanks, indeed looks like the correct thing is to close Body, even for HEAD requests.

I was originally going by this: https://stackoverflow.com/a/18601625/5821

> (...) If you don't need the body of the response, you should use Head instead of Get. Head doesn't require reading or closing the response body.

It's an 8 years old answer, so may be out of date. 

I think `resp.Body.Close()` needs to run only if there is no error. Otherwise, resp may be nil, and the code will panic.

hjr265:
@cuu508 You are right. I should have also added a return inside the `if` block.

And, the 8 year old answer is still relevant. (https://golang.org/src/net/http/client.go?s=20406:20458#L270)

Sorry for the noise.

PR #545, opened 2021-07-20
feat: manager role

### LINKED ISSUE #484: [FEATURE REQUEST] permission to add other members
It would be great to have an owner-like permission setting for team members, for them to add other members.
This would ease the management of projects which many members.

Possible permissions would then be:
- Owner
- Manager <-- NEW
- Member
- Read-Only

What do you think about that idea?

swoga:
This PR adds a manager permission as mentioned in the related issue.

The boolean field `Member.rw` is migrated to the enum field `Member.permission`.
Managers are allowed to invite/remove team members except themselves.

Closes #484

### DIFF
--- a/hc/accounts/forms.py
+++ b/hc/accounts/forms.py
@@ -6,7 +6,7 @@ from django import forms
 from django.core.exceptions import ValidationError
 from django.contrib.auth import authenticate
 from django.contrib.auth.models import User
-from hc.accounts.models import REPORT_CHOICES
+from hc.accounts.models import REPORT_CHOICES, Member
 from hc.api.models import TokenBucket
 import pytz
 
@@ -136,7 +136,7 @@ class ChangeEmailForm(forms.Form):
 
 class InviteTeamMemberForm(forms.Form):
     email = LowercaseEmailField(max_length=254)
-    rw = forms.BooleanField(required=False)
+    role = forms.ChoiceField(choices=Member.Role.choices)
 
 
 class RemoveTeamMemberForm(forms.Form):
--- a/hc/accounts/migrations/0043_add_role_manager.py
+++ b/hc/accounts/migrations/0043_add_role_manager.py
@@ -1,0 +1,17 @@
+from django.db import migrations, models
+from hc.accounts.models import Member
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('accounts', '0042_remove_member_rw'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='member',
+            name='role',
+            field=models.CharField(choices=Member.Role.choices, default=Member.Role.REGULAR, max_length=1),
+        ),
+    ]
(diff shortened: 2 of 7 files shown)


cuu508:
Awesome, thank you!

Before I can merge this, a few notes:

* READONLY, READWRITE and MANAGER look more like roles, not permissions to me. I would like to rename Member.permission to Member.role, and change the constant names in Python code: ROLE_READONLY, ROLE_MEMBER, ROLE_MANAGER. How does that sound?

* The PR adds Member.permission and removes Member.rw in a single migration. This will cause problems when doing rolling code deploys. The change would need to be done in several steps:

1. Add Member.role field (make it nullable so the old code works). Update application code to fill data in both Member.rw and Member.role, but treat Member.rw as the source of truth. Deploy, run migration.
2. Write a data migration that populates Member.role. Deploy, run migration.
3. Update application code to treat Member.role as the source of truth. Make Member.rw nullable. Deploy, run migration.
4. Remove the Member.rw field. Deploy, run migration.
5. Add support for the new "Manager" role.

I can do steps 1-4, and get the project to a state where the Member model has a "role" field with two possible values: "r" and "w". You could then rebase this PR and update it to add the third value: "m". Does that sound OK?

swoga:
> * READONLY, READWRITE and MANAGER look more like roles, not permissions to me. I would like to rename Member.permission to Member.role, and change the constant names in Python code: ROLE_READONLY, ROLE_MEMBER, ROLE_MANAGER. How does that sound?

Renaming `permission` to `role` is completely fine for me.
I considered calling it `MEMBER` instead of `READWRITE`, but decided on the latter to avoid confusion with `is_member` in some places and the `Member` model in general.

> I can do steps 1-4, and get the project to a state where the Member model has a "role" field with two possible values: "r" and "w". You could then rebase this PR and update it to add the third value: "m". Does that sound OK?

This would be great!
Do you mean splitting these migrations into multiple releases? I have never thought of rolling deployments.

cuu508:
> but decided on the latter to avoid confusion with is_member in some places and the Member model in general.

Ah, good point, that would be confusing. Let's do READONLY, REGULAR and MANAGER.

> Do you mean splitting these migrations into multiple releases? I have never thought of rolling deployments.

I'm thinking one release but multiple commits. I'm assuming most self-hosted Healthchecks instances are single-node, or can afford a couple minutes of dashboard not working during the upgrade. My main concern is the hosted service at https://healthchecks.io, a minute of downtime there would be noticed. But I routinely update it between releases, and can update commit-by-commit.

cuu508:
I added commits for steps 1-4. The current master now has a Member model with a role field. There's no new functionality yet, and no MANAGER role, I've just changed how the normal member / read-only member role is stored.

@swoga you should now be able to add the Manager role. I suspect it may be easier to copy-paste your changes from the PR rather than rebasing it, because there were a bunch of small changes in the test files.

swoga:
@cuu508, I updated my PR based on the current master.

cuu508:
Looks great, thank you!

PR #555, opened 2021-08-25
Fix db-data volume in docker-compose.yml

ypapouin:
Use data folder otherwise a second volume is created that becomes orphaned after a docker-compose down.
That means that, currently, after each up&down you start with a fresh database.


cuu508:
Hi @ypapouin, thank you for the PR!

I verified that it is like you say. But I don't understand what was wrong with the original compose file, and why this fix works. Do you have any tips?

Issue #594, opened 2022-01-05
Docker build failing because of requests package version conflict


renesas-brandon-hussey:
I was trying out a self-hosted version of healthchecks. I am not able to build the Docker Image. I get the following error:

```
ERROR: Cannot install requests 2.26.0 (from /wheels/requests-2.26.0-py2.py3-none-any.whl) and requests 2.27.0 (from /wheels/requests-2.27.0-py2.py3-none-any.whl) because these package 
versions have conflicting dependencies.                                                                                                                                                 
                                                                                                                                                                                        
The conflict is caused by:                                                                                                                                                              
    The user requested requests 2.26.0 (from /wheels/requests-2.26.0-py2.py3-none-any.whl)                                                                                              
    The user requested requests 2.27.0 (from /wheels/requests-2.27.0-py2.py3-none-any.whl)
```

I did a little digging. The requirement for requests v2.26.0 is coming from this project's requirements.txt.

https://github.com/healthchecks/healthchecks/blob/master/requirements.txt

The requirement for requests v2.27.0 seems to be coming from apprise.

```
RUN pip wheel --wheel-dir /wheels apprise uwsgi
```

For my testing, I just changed the version in this project to v2.27.0. Maybe consider specifying a version range?

cuu508:
Hey @renesas-brandon-hussey – thank you for the report!
I think I found an OK solution for this: fix the Dockerfile to download all wheels in one go. In that case, pip is clever enough to pick a single requests version, and not to collect (and later try install) two different versions.

I'll bump the requests version too to be on the latest version, but this change should prevent the same issue from happening when requests releases their next version.

PR #595, opened 2022-01-05
Fix requests package version conflict

renesas-brandon-hussey:
This is related to https://github.com/healthchecks/healthchecks/issues/594

Updating the package to v2.27.0 in requirements.txt failed about an hour after creating the original issue because v2.27.1 was released. For that reason I just changed it to >= v2.26.0. The docker image builds for me with this change.


cuu508:
Fixed in https://github.com/healthchecks/healthchecks/commit/d615cde23e63fdc4f980dbdb6edab93ce97c19c7

PR #597, opened 2022-01-13
Enable Django setting CSRF_TRUSTED_SOURCES

jormaj:
On my dockerized setup (using an Nginx reverse proxy) I got CSRF errors since the move to Django 4 (also see https://docs.djangoproject.com/en/4.0/ref/settings/#csrf-trusted-origins).
After some investiation I found out that this is caused by the fact that the proxy is ssl terminating (ie the proxy is reached through https), and the docker container is running on localhost (ie http).

The Django [csrf check](https://github.com/django/django/blob/3ff7f6cf07a722635d690785c31ac89484134bee/django/middleware/csrf.py#L275)
looks at the scheme of the request, so there's a mismatch between the scheme + host and the origin (ie http:// vs https://).
A workaround is to use CSRF_TRUSTED_SOURCES and set the origin there as a trusted origin.


cuu508:
Thanks for bringing this up, @jormaj!

What Dockerfile are you using? And what what WSGI server (gunicorn, uwsgi, ...) are you using?

I did a little digging and here's how it works on healthchecks.io:

* there is a TLS-terminating load balancer which sets X-Forwarded-Proto=http|https header. 
* I don't have CSRF_TRUSTED_SOURCES set
* I _also_ don't have [SECURE_PROXY_SSL_HEADER](https://docs.djangoproject.com/en/4.0/ref/settings/#secure-proxy-ssl-header) set
* but Django's `Request.is_secure()` still returns True. 
* Looks like this works because: 
    * UWSGI checks the `X-Forwarded-Proto` header and sets `wsgi.url_scheme` variable to "https" ([docs](https://uwsgi-docs.readthedocs.io/en/latest/Changelog-2.0.8.html#automatically-manage-http-x-forwarded-proto))
    * Django's WSGIRequests checks `wsgi.url_scheme` variable ([source](https://github.com/django/django/blob/main/django/core/handlers/wsgi.py#L91))

jormaj:
Hi, you're right, `X-Forwarded-Proto` did the trick! Many thanks, I learned something here.
I will close the pull request.

On a related note: I see that you included the above info to the docs, which I think is very useful for other users :)
I didn't fully understand one part there though:

* strips the X-Forwarded-Proto header from all incoming requests
* sets the X-Forwarded-Proto header to "https" only for requests that come via HTTPS

What I did (Nginx reverse proxy config) is the following:
`proxy_set_header X-Forwarded-Proto $scheme;`
which means the X-Forwarded-Proto header will be set to the protocol that was used by the original request (so if I'm using https://mywebsite.com it will be set to https, and vice versa)

I guess in this case the result is the same, but just forwarding the protocol that was used is more straightforward..?
Anyway, minor thing :)

cuu508:
@jormaj yes, setting `X-Forwarded-Proto` to nginx $scheme variable is fine. The important bit is to not allow the client control the `X-Forwarded-Proto` header. If the client sends this header, it *should not* be passed on as-is.

In the docs, I think if I had written

> * sets the X-Forwarded-Proto header to the nginx `$scheme` variable

it would be cleared to nginx users, but less clear to everyone else. If you have an idea how to word this better, please let me know! :-)

jormaj:
Ah ok, I see what you mean!
I was thinking more along the lines of:

> Set the X-Forwarded-Proto header to the protocol that was used in the original request. 
For example, using Nginx you can use the `$scheme` variable like so: `proxy_set_header X-Forwarded-Proto $scheme;`

Anyway, I think your explanation already helps plenty and should make sure anyone who is trying to make this work will know what to pay attention to :)

cuu508:
Thanks, I tweaked the wording a bit, and added the NGINX example.

PR #600, opened 2022-01-19
Adds description to pd template

vanugrah:


cuu508:
Thanks for the PR, @vanugrah! I think it's a good idea to pass the description to PagerDuty, but I think the "description" field is not the right place (despite the name!).

Looking at PagerDuty API docs, and at the UI, the "description" field is intended for a title or for a brief summary of the issue. It has a 1024 character limit. On the other hand, in Healthchecks the description field has no size limit.

PagerdDuty API has a [details field](https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTc4-send-a-v1-event), we could put additional data in there. I did a quick experiment:

image

I see two problems with this as-is:

* The fields are displayed in A-Z order, there doesn't seem to be a way to specify the display order. It would be nicer to show the "Last ping" and "Total pings" fields next to each other, for example.
* PagerDuty seems to render some extra whitespace (See how titles in the table don't line up with values. The screenshot is from Chrome, it's even worse in Firefox).

I asked PagerDuty support about both.

cuu508:
Created PagerDuty support forum topic: https://community.pagerduty.com/forum/t/custom-details-field-ordering-and-whitespace/3671

vanugrah:
@cuu508  I think that's a great idea. Love the use of custom fields. Thank you for looking into this and let me know if you need anything else from me.

cuu508:
PagerDuty support replied:

- they confirmed the extra whitespace in the "Custom Details" section is a bug, and they will work on it
- they will look into specifying the field order (but no promises)

I just pushed a commit that adds the extra fields in the PagerDuty payload.

PR #617, opened 2022-03-08
Prevent localization in non english language

egguy:
This allows to change the schedule of a check when using a non-English language.

When healthcheck is using a language like french, the total_seconds from the duration is formatted to the current locale.

This change `3600` to `3600,00` which prevent editing and raise an error 400. You can re specify each time the 2 values, but it's a lot of work.

This pull request prevents this behavior and allow you to edit directly.

### DIFF
--- a/templates/front/details.html
+++ b/templates/front/details.html
@@ -1,5 +1,5 @@
 {% extends "base.html" %}
-{% load compress humanize static hc_extras %}
+{% load compress humanize l10n static hc_extras %}
 
 {% block title %}{{ check|down_title }}{% endblock %}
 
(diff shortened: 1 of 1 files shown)


cuu508:
Thank you for reporting this issue, and for the PR!

I just pushed a fix that solves the problem slightly differently: output the `timeout` and `grace` values as integers not floats.

Before the fix, if a check has 1-minute duration, the template would contain:

    data-timeout="60.0"

Or, depending with a different locale, it could be:

    data-timeout="60,0"

There's no need for the decimal part, as the lowest period/grace granularity in Healthchecks is minutes. I added "floatformat" template filters to change the output to:

    data-timeout="60"

PR #622, opened 2022-03-22
Prettified matrix notifications

rwjack:
Just made the matrix notifications a bit prettier, since they were quite hard to notice among others in my healthcheck notifications matrix room.
IMHO this seems like a decent template, perhaps for other notification integrations?

Before:
image

After:
image


cuu508:
Thanks for the suggestion!

I think the red/green emojis are a good idea. They make the notifications easier to spot, even without the other changes. So I added them: https://github.com/healthchecks/healthchecks/commit/087ed0ac128aab5d2d34f540f65529b76afeddc5

PR #630, opened 2022-04-01
Add `hc_check_started` prometheus metrics

ajgon:
This is simple fix adds another metric to prometheus endpoint - `hc_check_in_progress`. Its purpose is to return `1` for checks which are currently in progress (marked as `started`, and not finished or failed). 

The main purpose are prometheus alerts. For example, in my case some of the cron tasks use 100% CPU, and it's normal - which means, I don't want to receive alerts from prometheus/alertmanager about CPU usage during this time. Thanks to this simple metric, I can add simple rule like `node_load5 > 8 and on() hc_check_in_progress{name="my-heavy-cron-job"} == 0` and never be bothered unnecessarily again :)

### DIFF
--- a/hc/front/views.py
+++ b/hc/front/views.py
@@ -1961,6 +1961,14 @@ def metrics(request, code, key):
             value = 0 if check.get_status() == "down" else 1
             yield TMPL % (esc(check.name), esc(check.tags), check.unique_key, value)
 
+        yield "\n"
+        yield "# HELP hc_check_started Whether the check is currently started in progress (1 for yes, 0 for no).\n"
+        yield "# TYPE hc_check_started gauge\n"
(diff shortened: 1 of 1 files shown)


cuu508:
Thanks for the PR!

How about calling this metric `hc_check_started` instead of `hc_check_in_progress`?

It's shorter, and "started" is already used elsewhere in the codebase, "in_progress" isn't. Also I think it would describe the check's state better: the check _was_ started, but no idea if it is still running or AWOL.

ajgon:
Hah! The funny thing is, I originally thought about naming it `hc_check_started`, but eventually settled on `_in_progress` for some reason :smile: 

I fixed the PR :+1:

PR #638, opened 2022-04-17
Rename .env to example.env

Firesphere:
Prevents customised .env files from interfering with git


cuu508:
What interference issue did you run into?

Firesphere:
If any changes in the `.env` happen in the repository (e.g. a new service is added, thus the `.env` file is changed), and they happen to be at or near an existing line (e.g. SLACK_TOKEN changes from line 10 to 11), Git will throw a conflict error, due to the local change (token added) and the remote changed (New line added).

I did not exactly run in to a problem as such, it's just a common issue when the `.env` itself is in the repository. Changing the `.env` to `example.env` avoids this, as it's fine to overwrite the example file, it won't break any existing values in the local `.env`

rbarros:
In many projects I see the `.env.example` pattern and not the suggested for you.

Firesphere:
> In many projects I see the `.env.example` pattern and not the suggested for you.

Yeah, it's a matter of who wrote it. Both are valid example files.

I personally prefer `example.env`, because `.env.example` would be a hidden file still after cloning, which possibly could confuse someone.

I'm not too fuzzed either way.

cuu508:
Thanks, OK, got it, the problem is merge conflicts. 
I prefer the {filename}.example naming scheme, we're already using it for `/hc/local_settings.py.example`.
Just renamed `/docker/.env` -> `/docker/.env.example`: https://github.com/healthchecks/healthchecks/commit/51d1b88a756536eecbd4570efc4f8e77bc9c546d

PR #660, opened 2022-05-31
Add last_start to Checks API

### LINKED ISSUE #633: API: Return "status" and "started" as separate fields
The web UI displays "up + started" and "down + started" states by showing a an animated progress spinner under the "up" or "down" icon:

image

In comparison, the API currently cannot express the "up + started" and "down + started" states. The API returns a single "status" field with the possible values: `new`, `started`, `up`, `grace`, `down`, and `paused`. Therefore custom dashboards are disadvantaged compared to the standard web UI (see https://github.com/healthchecks/dashboard/issues/8). 

Update API to return two fields:

* status = new|up|grace|down|paused
* started = true|false

This would be backwards-incompatible change so would probably have to go under /api/**v2**/.

crogi93:
Adds `last_start` field to `GET /v1/checks/{id-or-slug}`.

Fixes #633

### DIFF
--- a/hc/api/models.py
+++ b/hc/api/models.py
@@ -252,6 +252,7 @@ class Check(models.Model):
             "grace": int(self.grace.total_seconds()),
             "n_pings": self.n_pings,
             "status": self.get_status(with_started=True),
+            "last_start": self.last_start,
             "last_ping": isostring(self.last_ping),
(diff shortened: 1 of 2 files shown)


cuu508:
Thanks for the PR!

A problem with this is it isn't able to report "up and started" and "down and started" states.

PR #661, opened 2022-06-02
Expose subject and subject_fail via API

### LINKED ISSUE #659: Update email keywords via API
Subject and subject_fail aren't exposed via the API:

https://github.com/healthchecks/healthchecks/blob/master/hc/api/models.py#L75

https://github.com/healthchecks/healthchecks/blob/master/hc/api/views.py#L97

Looks like a pretty simple change I could throw together a PR for. Are there any docs/tests/other things that would also need to be updated?

TFenby:
Closes #659


TFenby:
Do I need to update the api.html file as well or is that generated somehow?

cuu508:
Thanks for the PR!

The .html files are generated from .md files with a management command –

    ./manage.py render_docs

I've done this, plus a few other tweaks, and I think we're all set now.

PR #666, opened 2022-06-10
Revised text email template for being more informative.

arowalnas:
Hello everyone,

for my own needs, I've added more outputs to the text-based email template. Perhaps this revision will be useful for others as well. However, I haven't been able to avoid the superfluous line breaks since I'm a pure Python/Django beginner. Does anyone have an idea how I could add this improvement?

Greetings,
Walid

### DIFF
--- a/templates/emails/alert-body-text.html
+++ b/templates/emails/alert-body-text.html
@@ -1,13 +1,110 @@
-{% load hc_extras %}
-The check "{{ check.name_then_code }}" has gone {{ check.status }}.
+{% load hc_extras humanize %}
+
+Dear User,
+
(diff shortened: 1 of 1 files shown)


cuu508:
Thanks for the suggestion!

I just pushed a revised email template: https://github.com/healthchecks/healthchecks/commit/f2ce7aed8e408716b63e7eca95410e2262da116e

> However, I haven't been able to avoid the superfluous line breaks since I'm a pure Python/Django beginner. Does anyone have an idea how I could add this improvement?

I've struggled with withespace control in Django templates as well, my solution was `{% linemode %}` and `{% line %}` template tags: https://github.com/healthchecks/healthchecks/blob/master/hc/front/templatetags/linemode.py

PR #669, opened 2022-06-15
Add support for multiple RCPT TO in incoming email

pesca-chingolo:
Found a particular scenario where the software (Veeam Agent) sending the email notifications that also have to be processed as pings, allows to send to one fixed recipient address only.
If I forward these emails, from that address to the multiple checks/recipients that I need (filtering via subject in Healthchecks), it sends them in one connection using multiple RCPT TOs, and Healthchecks just processes the first one.

This little change allows it to process multiple RCPT addresses and may be useful for this particular case and others.

### DIFF
--- a/hc/api/management/commands/smtpd.py
+++ b/hc/api/management/commands/smtpd.py
@@ -61,8 +61,9 @@ class Listener(SMTPServer):
         # get a new db connection in case the old one has timed out:
         connections.close_all()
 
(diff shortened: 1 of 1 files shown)


cuu508:
Thanks for the PR!
Could you also add a testcase with multiple recipients [here](https://github.com/healthchecks/healthchecks/blob/master/hc/api/tests/test_smtpd.py)?
Thanks!

PR #685, opened 2022-08-02
Add support for SMTP with implicit TLS

facorazza:
I've added `EMAIL_USE_SSL` to the environment variables handled by `settings.py` and changed the default SMTP port to `465` which is the recommended one over `587`. I've also set their default value to `False` as they can't be both `True` at the same time. A user who wants to use either one has to specify the correct one.

I've also updated the `README.md` with the new configuration setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,13 +110,23 @@ Healthchecks must be able to send email messages, so it can send out login
 links and alerts to users. Specify your SMTP credentials using the following
 environment variables:
 
-```python
-EMAIL_HOST = "your-smtp-server-here.com"
-EMAIL_PORT = 587
-EMAIL_HOST_USER = "smtp-username"
-EMAIL_HOST_PASSWORD = "smtp-password"
-EMAIL_USE_TLS = True
-```
+- Implicit TLS (*recommended*):
+    ```python
+    EMAIL_HOST = "your-smtp-server-here.com"
+    EMAIL_PORT = 465
+    EMAIL_HOST_USER = "smtp-username"
+    EMAIL_HOST_PASSWORD = "smtp-password"
+    EMAIL_USE_SSL = True
+    ```
+
+- Explicit TLS:
+    ```python
+    EMAIL_HOST = "your-smtp-server-here.com"
+    EMAIL_PORT = 587
+    EMAIL_HOST_USER = "smtp-username"
+    EMAIL_HOST_PASSWORD = "smtp-password"
+    EMAIL_USE_TLS = True
+    ```
 
 For more information, have a look at Django documentation,
 [Sending Email](https://docs.djangoproject.com/en/1.10/topics/email/) section.
--- a/README.md
+++ b/README.md
@@ -116,8 +116,11 @@ environment variables:
     EMAIL_PORT = 465
     EMAIL_HOST_USER = "smtp-username"
     EMAIL_HOST_PASSWORD = "smtp-password"
+    EMAIL_USE_TLS = False
     EMAIL_USE_SSL = True
     ```
+
+    Port 465 should be the preferred method according to [RFC8314 Section 3.3: Implicit TLS for SMTP Submission](https://tools.ietf.org/html/rfc8314#section-3.3). Be sure to use a TLS certificate and not an SSL one.
 
 - Explicit TLS:
     ```python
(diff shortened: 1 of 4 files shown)


cuu508:
Thanks for the PR! But is 465 and implicit TLS indeed recommended? A quick web search seems to suggest explicit TLS is the recommended one.

From https://sendgrid.com/blog/whats-the-difference-between-ports-465-and-587/:

> **Which should I use?**
> Tl;dr Use port 587 if you can, 465 if you can’t, 25 if you must.
> Port 587 is technically correct, the best kind of correct. However, many ESPs have adopted implicit TLS on port 465. While you can send email over port 25 and 2525, it’s much more secure to have the messages encrypted. This makes port 587 the preferred option for sending, with port 465 as a close second.

From https://stackoverflow.com/questions/15796530/what-is-the-difference-between-ports-465-and-587:

> Historically, port 465 was initially planned for the SMTPS encryption and authentication “wrapper” over SMTP, but it was quickly deprecated (within months, and over 15 years ago) in favor of STARTTLS over SMTP (RFC 3207). Despite that fact, there are probably many servers that support the deprecated protocol wrapper, primarily to support older clients that implemented SMTPS. Unless you need to support such older clients, SMTPS and its use on port 465 should remain nothing more than an historical footnote.

From https://kinsta.com/blog/smtp-port/:

> While many providers still do support port 465 for SMTP, it’s no longer an accepted standard and you should always try to use ports 587 and 2525 before using port 465.

From https://mailtrap.io/blog/smtp-ports-25-465-587-used-for/:

> Whenever you’re given a choice, you should use 587 as the default SMTP port. It comes with TLS security and is the officially recognized port for email submission.

From https://www.cloudflare.com/learning/email-security/what-is-smtp/:

> Port 465 was once designated for use by SMTP with [Secure Sockets Layer (SSL)](https://www.cloudflare.com/learning/ssl/what-is-ssl/) encryption. But SSL was replaced by [Transport Layer Security (TLS)](https://www.cloudflare.com/learning/ssl/transport-layer-security-tls/), and modern email systems therefore do not use this port. It only appears in legacy (outdated) systems.

facorazza:
Indeed, the use of SSL certificate is discouraged over the use of TLS. Historically, SSL certificates have usually been used on port 465 and as you correctly pointed out port 587 should be preferred instead as that provides the capability to **negotiate** the use of TLS (encryption is not necessarily enforced even after the initial plain text exchange). However, port 465 has been "recently" (2018) designated to the use of **implicit TLS** and not SSL, so that encryption is actually enforced for all communications between client and server.

Some references:
- [docker mailserver docs](https://docker-mailserver.github.io/docker-mailserver/edge/config/security/understanding-the-ports/)
- [RFC8314 Section 3.3: Implicit TLS for SMTP Submission](https://tools.ietf.org/html/rfc8314#section-3.3)

Regarding this PR, in the [Django's documentation](https://docs.djangoproject.com/en/4.0/ref/settings/#std-setting-EMAIL_USE_SSL), `EMAIL_USE_SSL` is used to enable "implicit TLS" so I suppose the use of SSL vs TLS comes down to the server using an SSL or a TLS certificate. For this reason, I would keep the PR as it is and probably state that a TLS certificate should be preferred over an SSL one even though I think that should be obvious by now.

cuu508:
Thanks for the pointers, interesting! So the bits I quoted are all out of date, and not best practice any more?

This PR changes the defaults in settings.py:

* `EMAIL_PORT` from 587 to 465
* `EMAIL_USE_TLS` from True to False

This can break existing installations that rely on the default values. I would prefer to leave the defaults as-is. Can you update the PR to add support for the EMAIL_USE_SSL env var, but leave the defaults unchanged?

Please also add `EMAIL_USE_SSL` in `/templates/docs/self_hosted_configuration.md`.

facorazza:
@cuu508 Getting back to this PR, since changing the port is still a breaking change I suggest we update `EMAIL_USE_SSL` and `EMAIL_USE_TLS` as well and merge the changes when releasing a major update (3.0.0).

cuu508:
I'm happy to merge a change that adds support for `EMAIL_USE_SSL` environment variable, but does not make any changes to the current defaults.

DerDanilo:
> I'm happy to merge a change that adds support for `EMAIL_USE_SSL` environment variable, but does not make any changes to the current defaults.

Was looking for this. There are a few servers that require SSL instead of TLS for SMTP connections. Thanks!

facorazza:
@cuu508 The PR is ready

PR #696, opened 2022-08-16
Update Dockerfile to require Lib cURL OpenSSL

Firesphere:
Updated requirements


cuu508:
Thanks for the PR! As it happens, I already made almost the same changes in https://github.com/healthchecks/healthchecks/commit/5baa8a53deaa067a18fba80725cdc6e908f42fc4 :-)

PR #708, opened 2022-09-24
Webhooks support for $BODY placeholder

bdd:
Adds support for sending ping body contents with webhooks.
Only allowed in the 'body' template of the webhook configuration.

Currently only e-mail transport supports including the ping body in the notification.

Implementing this for webhooks seemed like the right place for maximum flexibility.

Example uses:
* Use Amazon API Gateway to receive and stash a copy of failure log on your own S3 account. (Inspired by discussion at https://github.com/bdd/runitor/issues/60)
* Use IFTTT, Zapier, n8n, etc. to receive the full context of a check status change.

### DIFF
--- a/hc/api/tests/test_notify_webhook.py
+++ b/hc/api/tests/test_notify_webhook.py
@@ -5,7 +5,7 @@ import json
 from unittest.mock import patch
 
 from django.utils.timezone import now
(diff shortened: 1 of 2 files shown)


cuu508:
Looks good, thanks for the PR, @bdd!

PR #711, opened 2022-10-03
Update  German django.po

Piccar:
Fixed typo and made  the translation sound more natural.


Piccar:
The German translation sounds very strict and formally. Is this intentional?

cuu508:
Thanks for looking into improving the translations!

> The German translation sounds very strict and formally. Is this intentional?

I would prefer a similar style as in English – neutral, not overly formal, not overly casual.

This was a contributed translation, my German is too poor to be able to assess its quality :-/

In any case, there has been only one page set up for translation – the "Welcome" page. A while ago [I removed the "Welcome" page](https://github.com/healthchecks/healthchecks/commit/9e36eb5fcc6f74213b7efb95c3cce924c6480a30) so most of the translated strings are obsolete now.

PR #714, opened 2022-10-11
Remove port portion of remote_addr to fix Azure deployments

cam-barts:
In an Azure deployment, there were issues saving ping objects to the database because the addresses appear to come in with ports attached to them (123.123.123.123:5314), which breaks the [GenericIPAddressField](https://docs.djangoproject.com/en/4.1/ref/forms/fields/#genericipaddressfield) constraint on the [Ping model](https://github.com/healthchecks/healthchecks/blob/11997e75c8b82981537234f236e0abb878afeeee/hc/api/models.py#L497). This had the following results: 
- Pings not actually saving to the database
- Hitting ping endpoints would result in 500 status codes
- Check's status would change, but no logs would appear on the page

I was able to make this change locally and push a container based on it into Azure and the fix works.

### DIFF
--- a/hc/api/models.py
+++ b/hc/api/models.py
@@ -379,6 +379,8 @@ class Check(models.Model):
             ping.kind = action
 
         ping.remote_addr = remote_addr
+        if ':' in remote_addr:
+            ping.remote_addr = remote_addr.split(':')[0] # If the address contains a port (like in Azure), remove the port portion
         ping.scheme = scheme
         ping.method = method
         # If User-Agent is longer than 200 characters, truncate it:
--- a/hc/api/models.py
+++ b/hc/api/models.py
@@ -379,7 +379,7 @@ class Check(models.Model):
             ping.kind = action
 
         ping.remote_addr = remote_addr
-        if ':' in remote_addr:
+        if ':' in remote_addr and '.' in remote_addr:
             ping.remote_addr = remote_addr.split(':')[0] # If the address contains a port (like in Azure), remove the port portion
         ping.scheme = scheme
         ping.method = method


cuu508:
Thanks for the PR. A couple questions:

* What would be the impact on IPv6 client addresses?
* In the Azure deployment, are the 123.123.123.123:5314 values appearing in the X-Forwarded-For  request header? Which component (Azure Load Balancer, Azure Gateway, some reverse proxy, ...) is setting them? Is this behavior (putting ip:port instead of just ip in X-Forwarded-For) documented somewhere?

cam-barts:
Ooh, I hadn't even thought about IPv6. Perhaps I can put another clause in the gaurd that checks that the address has a period in it to differentiate IPv4. 

As for Azure, I'm not very certain where that value comes from. I found the issue by running the application in debug mode and taking note of the errors, which all had to do with writing the value not being parsed as an inet. After some digging it appeared that was coming from the Django model, as I manually set the column type in the database to text and it didn't resolve the initial issue. 

I am running the [linuxserver.io](https://github.com/linuxserver/docker-healthchecks) docker image on a single instance [Azure App Service](https://azure.microsoft.com/en-in/products/app-service/) pulling directly from docker hub and using postgres as the data store. I don't have a managed load balancer, gateway, or reverse proxy in front of it, and as such there isn't much of a way for me to change headers that I'm aware of.

cam-barts:
Looking more into it, my best guess is that it is indeed coming from the X-Forwarded-For header that the. I did find this [MSDN post](https://social.msdn.microsoft.com/Forums/en-US/e2357255-ccf4-48bb-bf4a-dd986f84b437/how-to-remove-port-from-xforwardedfor-header-in-azure-webapp?forum=windowsazurewebsitespreview) from 2018 (which links to a [un-navigable feature request](https://feedback.azure.com/forums/217313-networking/suggestions/20220997-support-for-dropping-port-out-of-x-forwarded-for-h)) that indicates that at the app service level it isn't possible to change that header. This [stack overflow response](https://stackoverflow.com/questions/50889876/how-to-remove-port-from-x-forwarded-for-header-in-azure-webapp) to a similar question (likely by the same person) suggests putting an azure gateway resource in front of the application. 

The issue I run into is that it seems like overkill to [pay for](https://azure.microsoft.com/en-us/pricing/details/application-gateway/) another resource to put in front of my web app just to format this header. My use case is such that I wouldn't really get any additional benefit out of the gateway.

Issue #319, opened 2020-01-15
Idempotence key for ping requests


cuu508:
When curl (another HTTP client) retries failed requests we can end up with multiple logged pings:

https://twitter.com/epikhinm/status/1217147491986133002

Look into making ping requests idempotent.

abandholm:
Maybe something simple like (assuming a shell like Bash):

`curl -fsS --retry 3 -H "HC-Id: $RANDOM" https://hc-ping.com/your-uuid-here`

Bash will add a random number, but curl will use the same value on all 3 calls -- server discards duplicates.

People who want idempotency include the id-header.

cuu508:
If there is idempotence key, it could be also used to measure job execution times for concurrent / overlapping jobs.

* 00:00 Job A starts (idempotence key "A")
* 00:10 Job B starts (idempotence key "B")
* 00:20 Job A finishes (based on the idempotence key "A", we can calculate the run time – 20 minutes)
* 00:25 Job B finished (based on the idempotence key "B", we can calculate the run time – 15 minutes)

PR #720, opened 2022-10-24
Added duration to ping details 

seidnerj:
This is useful on devices that have a small screen, since the duration cannot be seen in the main view. With this,  one can see the duration in the ping's details.


seidnerj:
I did, but apparently I did not do such a good job. I will re-check it.

Thanks!

seidnerj:
Fixed. Serves me right for not properly testing this, sorry.

seidnerj:
By the way, seems like a bit of waste to recalculate the duration every time - maybe it should be saved as a field in the database instead when the ping is received. This of course would assume that a ping's "history" does not change retroactively.

cuu508:
> By the way, seems like a bit of waste to recalculate the duration every time - maybe it should be saved as a field in the database instead when the ping is received. This of course would assume that a ping's "history" does not change retroactively.

It's a tradeoff – we would use more space (and slightly more IO) to store the precalculated duration in the `api_ping` table, but would avoid recalculating it at display time. I think currently the tradeoff is not worth it:

* The `api_ping` table is the biggest table in the database, and is constantly being modified. Database write IOPS is a limited resource.
* Most pings get written, never looked at by human operator, and cleaned up later. For those pings, storing the duration is wasted work and space.

On the other hand,

* After this PR, we will have two places where we calculate the duration. If the duration is calculated once and stored, the calculation is simpler (`delta = now() - check.last_start`) and only needs to happen in a single place.
* Now thinking about it, there may be a bug in the events list, where the duration for the very last item in the events list  is not displayed. If the duration is calculated once and stored, this bug would go away.

For now, let's stick with calculating the duration on the fly.

---

Feedback on the PR: good work, AFAICT it works correctly now! I have a few suggestions though :-)

```
pings = self.owner.ping_set.filter(created__gte=self.created - MAX_DELTA).order_by("-id")
```

It's a great idea to fetch pings from only MAX_DELTA in the past. It can be optimized to also exclude any pings that come *after* our current ping:

```
q = self.owner.ping_set
# only look backwards:
q = q.filter(id__lt=self.id)
# don't look further than MAX_DELTA in the past
q = q.filter(created__gte=self.created - MAX_DELTA)
```

And, if we are only interested in the duration of a single ping, the for loop can be optimized to go backwards from the ping until we hit a "start" event:

```
    def get_delta(self) -> timedelta | None:
        # return early if this is not a success or failure ping
        if self.kind not in (None, "", "fail"):
            return None

        q = self.owner.ping_set
        # only look backwards:
        q = q.filter(id__lt=self.id)
        # don't look further than MAX_DELTA in the past
        q = q.filter(created__gte=self.created - MAX_DELTA)

        # look for the nearest "start" event
        for ping in q.order_by("-id"):
            if ping.kind == "start":
                return self.created - ping.created

        return None
```

In `hc.front.views.ping_details` you have:

```
    delta = ping.get_delta()
    if delta is not None:
        setattr(ping, "delta", delta)
```

How about decorating `Ping.get_delta()` method with the `@cached_property` decorator and using it directly in the template? Perhaps also rename it to `duration`? So the template could do `{% if ping.duration %} ... {% endif %}`.

And, last but not least, please add test cases that exercise the new functionality.

seidnerj:
Great suggestions! I've implemented all of them in the above commit.

Re adding a test case, where would be the best location to put add a relevant test?

seidnerj:
By the way, another thought - the duration's calculation is wrong in the case where you have two overlapping runs, for example:

13:00 - Run 1 starts 
13:01 - Run 2 starts
13:05 - Run 1 ends
13:06 - Run 2 ends

As I understand the way the calculation is done today, run 2's duration would be correct, i.e. 5 minutes, but run 1's duration would incorrectly be calculated as as 4 minutes even though in reality it is 5 minutes.

cuu508:
Awesome!

I'd add new test case(s) in `hc.front.tests.test_ping_details`. The tests would create Ping instances and test that the duration is displayed in the "Ping Details" dialog as expected.

cuu508:
Yes, if you have overlapping runs then you cannot trust the calculated durations. One idea to solve that would be to introduce optional idempotence keys when pinging: https://github.com/healthchecks/healthchecks/issues/319#issuecomment-818458171

seidnerj:
> Yes, if you have overlapping runs then you cannot trust the calculated durations. One idea to solve that would be to introduce optional idempotence keys when pinging: [#319 (comment)](https://github.com/healthchecks/healthchecks/issues/319#issuecomment-818458171)

Yep, but that would already be quite a big change... :|

seidnerj:
> Awesome!
> 
> I'd add new test case(s) in `hc.front.tests.test_ping_details`. The tests would create Ping instances and test that the duration is displayed in the "Ping Details" dialog as expected.

Great! so anything I need to do or we're set?

cuu508:
Sorry, I may have worded it confusingly – can you add test cases in `hc.front.tests.test_ping_details`?

seidnerj:
Sure, will do.

seidnerj:
Done.

cuu508:
Thanks for all the work, @seidnerj!

I reviewed and tested the changes, all looks good except one efficiency regression which I'll fix. The regression: the code in `hc.front.views.ping_details` calculates durations and sets them on the `Ping.duration` field, overriding the cached property. But, it only does this for pings where the duration can be calculated. If there's a streak of "success" pings with no "start" pings between, the loop will not set the `Ping.duration` field for most of them. So, when rendering the template, the property code will run and cause an additional SQL query for each ping. The fix is, in `hc.front.views.ping_details`, to set the `Ping.duration` field for each and every ping.

seidnerj:
Thank you!

PR #722, opened 2022-11-02
Added support for a "run id" parameter

seidnerj:
This can be used for properly calculating duration of overlapping runs and (in the future) idempotence.


seidnerj:
There are 2 tests that fail, but it does not seem to be related to the changes I've made in this pull request?

seidnerj:
Never mind, I realized what was the issue. This should now be ready to merge.

seidnerj:
Helps #319

cuu508:
Awesome, thanks for the PR! 

FYI –  I'm a little busy right now, may take a couple days before I can review.

seidnerj:
No worries

cuu508:
Initial comments:

* I like the idea of passing the run id as a query parameter (`?rid=...`). Easy to use with any HTTP client, run ids visible in access logs.
* In the model, the new `rid` field definition needs `null=True`. Without it, the migration will be expensive on a large `api_ping` table.
* `rid` is currently `varchar(36)`. I'm assuming the idea is to allocate enough space for UUID string to fit, but accept anything that is 36 characters or shorter. Since we're adding a new feature, and don't have backwards-compatibility concerns (yet), I'd prefer to be stricter here and mandate a specific datatype here. And use a more specific database field, either IntegerField (4 bytes) or UUIDField (16 bytes). With integer, in shell scripts users can generate the run id using `$RANDOM`, and UUIDs can be generated with `uuidgen`. With a random integer the risk of collisions is of course higher.
* In the worst-case scenario `_get_events` can generate N queries. Imagine a case where the client submits only "success" events, each with an unique run id. `_get_events` will leave the `duration` field untouched, and the template will fall back to calling `Ping.duration()` for every ping. I think we will need to have protection against this. Perhaps something like, if duration calculation uses too many queries, then don't show durations at all. But that can perhaps be a separate PR.

seidnerj:
Agreed on all points. Re you past point, the N queries, an alternative to not showing durations at all would be to gradually load durations in chunks of say 10, if N is greater than a certain threshold. That way the load is limited. This can be configurable behavior.

At any rate, I think this can be a separate PR.

seidnerj:
Assuming, I under your point, another thing that could be done is that _get_events will not leave the duration field empty but rather mark it with some value that indicates a calculation was attempted but "failed", this could then be ignored during rendering and the aforementioned queries would not occur.

cuu508:
The three migrations can be combined into one:

1. on your dev system, "un-apply" the new migrations: `./manage.py migrate api 0093`
2. delete the migration files for 0094, 0095, and 0096
3. run `./manage.py makemigrations` again and it should create a single migration

cuu508:
A couple more notes:

1. the `_get_events` function can get away with using just one dictionary for tracking start timestamps:

```python
    # Optimization: the template will access Ping.duration, which would generate a
    # SQL query per displayed ping. Since we've already fetched a list of pings,
    # for some of them we can calculate durations more efficiently, without causing
    # additional SQL queries:
    starts = {}
    for ping in reversed(pings):
        if ping.kind == "start":
            starts[ping.rid] = ping.created
        elif ping.kind in (None, "", "fail"):
            if ping.rid in starts:
                # If we've seen a start/success/failure event for this run id,
                # make sure we do not fall back to Ping.duration()
                ping.duration = None

            start = starts.get(ping.rid)
            if start:
                if ping.created - start < MAX_DURATION:
                    ping.duration = ping.created - start

            starts[ping.rid] = None
```

The dictionary keys are UUIDs (run ids). Values:

* If a value for a given dictionary key is a datetime, that's the timestamp of the start event for that run id.
* If a value for a given dictionary key is None, it means the previous event for that run id was "success" or "fail".
* If the dictionary does not contain a key, it means we've seen no "start", "success" or "fail" events for that run id, and will need to fall back to the `Ping.duration()` call.

2. In the model, Ping.rid does not need the `blank=True` keyword:

```
rid = models.UUIDField(null=True)
```

3. We need to handle cases when user supplies something other than UUID in the rid parameter, and corresponding test cases.

seidnerj:
Done.

cuu508:
Awesome, looks good!

There are two more changes I want to do, to lock down the initial implementation a little more:

1. Limit the UUID formats accepted in the rid parameter to only this form `c0b921e6-5f47-6ec2-c793-f13f4cc04435` (and reject e.g. `{c0b921e6-5f47-6ec2-c793-f13f4cc04435}`, `c0b921e65f476ec2c793f13f4cc04435`. For consistency could perhaps use the same regex that Django's UUID URL converter uses: https://github.com/django/django/blob/5f09ab8c30050bbd076a9b27fb135d030c06ab75/django/urls/converters.py#L26
Rationale: I want to declare a single format that will be supported, and not worry about any unintended consequences of supporting the alternative formats that uuid.UUID happens to support. Fewer things to document too.

2. Read the rid parameter from request.GET only, don't read from request.POST. 
Rationale: simpler implementation, no potential for confusion when somebody sends rid both in the query string and the POST body, and, again, fewer things to document. 

If there's demand and use cases, there would be no problem adding these bits back later. But it would be tricky to remove them once they're out in the wild.

Would you be OK to make these changes? I can also take over from here if you wish.

seidnerj:
No worries, re 1 - I'll check what's the best way to implement this.

Re 2, I was under the impression that request.POST.get("rid") would fetch the the rid argument from the query string of a POST request, and is not related to the body of the request, is my impression incorrect?

cuu508:
> Re 2, I was under the impression that request.POST.get("rid") would fetch the the rid argument from the query string of a POST request, and is not related to the body of the request, is my impression incorrect?

The query string parameters are in `request.GET`, even for HTTP POST requests. `request.POST` contains form data.

seidnerj:
Got it, okay. This was my original intention anyways. I will make the appropriate changes.

seidnerj:
Fixed.

cuu508:
Thanks for the work, @seidnerj !

seidnerj:
My pleasure, when do you expect this to be deployed on http://healthchecks.io? would be nice to get a heads up once it is up. Thanks!

cuu508:
I'll add protection for the n queries problem, and then it should be good to go. Should not take more than a couple days. I'll comment here once it's up.

seidnerj:
Awesome. Thanks!

cuu508:
... and it's deployed. (There's no mention of run ID in the documentation yet, for now it is an undocumented feature :-))

seidnerj:
😁🙏👍

seidnerj:
Just deployed this to my processes - works great, thanks! :)

PR #735, opened 2022-11-17
Add body to Slack Notification

s256:
We were missing the body / payload of the last request triggering a notification. 
Hence we added the body to the template, if it exists.


cuu508:
Thanks for the PR, @s256!

s256:
Of course! Great project. Using it on a daily basis :)

PR #737, opened 2022-11-23
Allow to retrieve a ping's body

lablans:
This PR adds a new query parameter to the `/api/v1/checks/<uuid>/pings/` API endpoint. If set to true, returned pings will contain their respective request bodies (if present). See the added documentation for more information.


cuu508:
The response can contain 1000 pings, and a ping body can be up to 10KB in size, so the total response size could be 10MB, and require 1000 S3 GetObject calls to produce.

What use case do you have in mind for this?

lablans:
Use-Case: First of all, there should be an API to retrieve anything that could also be retrieved via a web browser. In my case, I'm using a self-hosted instance as a combined monitoring and logging solution (by pinging important events with `/log`), which is very helpful to see both the status and why it is degraded. I'm also analyzing these log entries by fetching them with a python script and acting upon their content. These are servers we don't have direct access to, so it's really helpful to have this information available centrally.

Thanks for informing me about S3 calls. This is a non-issue for self-hosted instances but I realize this could become problematic for the official instance. If we're concerned about result lists getting too big, here's two ideas:
* Introduce paging, e.g. `since=<timestamp>` or `maxage=<seconds>` or `page=1&pagesize=10`. This could be a useful feature on its own since may use-cases will only be interested in n=100 or maxage=3600 (or similar) pings anyway.
* Move into a separate deliberate API call, e.g. `/api/v1/checks/<uuid>/pings/<n>/body`

I'd be happy to implement the latter and update this PR, if you like.

cuu508:
> Move into a separate deliberate API call, e.g. `/api/v1/checks/<uuid>/pings/<n>/body`

Sounds good, I'd accept a PR for that!

lablans:
> > Move into a separate deliberate API call, e.g. `/api/v1/checks/<uuid>/pings/<n>/body`
> 
> Sounds good, I'd accept a PR for that!

Here you go. For the sake of RESTful-ness, I added a link to the body (`body_url`) to each Ping. Feel free to remove if you are concerned about bandwidth.

hc/api/views.py
```diff
@@ -424,6 +424,25 @@ def pings(request, code):
     return JsonResponse({"pings": [p.to_dict() for p in pings]})
 
 
+@cors("GET")
+@csrf_exempt
+@validate_json()
```
cuu508, line 429: not needed here, I think

hc/api/models.py
```diff
@@ -541,6 +541,18 @@ class Ping(models.Model):
...
     def to_dict(self) -> PingDict:
+        if self.has_body():
+            body_url = (
+                settings.SITE_ROOT
+                + "/api/v1/checks/"
+                + str(self.owner.code)
+                + "/pings/"
+                + str(self.n)
+                + "/body/"
+            )
```
cuu508, line 552: I would prefer to construct the URL using `django.urls.reverse()`.

hc/api/views.py
```diff
@@ -424,6 +424,25 @@ def pings(request, code):
...
+@authorize
+def ping_body(request, code, n):
+    check = get_object_or_404(Check, code=code)
+    if check.project_id != request.project.id:
+        return HttpResponseForbidden()
+
```
cuu508, line 435: We should also check here if `n` is not too far back.  Something like (untested):

```python
profile = Profile.objects.get(user__project=request.project)
threshold = check.n_pings - profile.ping_log_limit
if n < threshold: 
    raise Http404()
```

cuu508 (commented):
Thanks for updating the PR! Some comments below, and could you also please add test cases for the new view?

hc/api/views.py
lablans: Corrected in 4824c89a

hc/api/models.py
lablans: Fixed in https://github.com/healthchecks/healthchecks/pull/737/commits/f9cdf9ec2a716bd86b007824e7bfdbc6aa2ad1fe

hc/api/views.py
lablans: Done in https://github.com/healthchecks/healthchecks/pull/737/commits/4a9758054e223b2dcd5c20a4a2cdf04fdeb7d246

lablans:
> Thanks for updating the PR! Some comments below, and could you also please add test cases for the new view?

All done -- and test added in https://github.com/healthchecks/healthchecks/pull/737/commits/f720f7e941fd4a95fb27b323ef411bc6f918ccad.

This was my first Python in years and my first time in Django. A nice, well-organized project to learn from. Thanks!

PR #783, opened 2023-01-26
Add body to Telegram notifications

f-f:
As per title, inspired by #735

### DIFF
--- a/hc/api/tests/test_notify_telegram.py
+++ b/hc/api/tests/test_notify_telegram.py
@@ -168,13 +168,14 @@ class NotifyTelegramTestCase(BaseTestCase):
 
         args, kwargs = mock_post.call_args
         payload = kwargs["json"]
+        self.assertIn("Last Ping Body:", payload["text"])
         self.assertIn("Hello World", payload["text"])
 
     @patch("hc.api.transports.curl.request")
     def test_it_shows_truncated_last_ping_body(self, mock_post):
         mock_post.return_value.status_code = 200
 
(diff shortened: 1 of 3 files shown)


f-f:
Hi @cuu508, I noticed that #785 (a very similar change) was merged but this one was not, is there anything I could do to help move it forward?

cuu508:
Hi @f-f, thanks for the PR! The next step is for me to test the changes. I hope to get around to it this week.

f-f:
Sounds great, thanks!

cuu508:
Just reviewed and tested the PR. Here are the things I'd like fixed or changed:

* Fix tests – some of the new test cases currently don't pass
* The telegram_message.html template does not output the ping body. The template uses the `{% line %}....{% endline %}` tags for precise whitespace control, and the body is not enclosed in these tags. (see the next point for an example on how to fix)
* The Telegram integration uses the "html" parsing mode, so Markdown syntax will not work:

image

use `<pre>` tags instead like so:


```
{% if body %}
{% line %}<pre>{{ body }}</pre>{% endline %}
{% endif %}
```

Result:

image

* The other integrations truncate long body at 1000 characters. For the sake of consistency, I'd like the Telegram integration to do the same.
* I think the pre section needs a label ("Last Ping Body"), otherwise, for some users, it may not be clear what the preformatted content is.

f-f:
Thanks for the review @cuu508! I pushed a commit that should address all the feedback.

Issue #784, opened 2023-01-31
[Feature] Add ability to send RESPONSE_BODY in webhook integration


boopzz:
We currently use Mattermost as our messaging system and use it for simple alerting also. Currently the Mattermost integration works but wont give the response body in a message. To give some more information with our scripts we send an error message in the body of our failed healthcheck which we'd like sent with the alert. I can see in the docs where we can use a webhook instead for the alert to Mattermost with some other placeholders.

It would be really cool to have another placeholder called $BODY or $RESPONSE_BODY that outputs the response body sent with the alert. Arguably not everyone would want this, especially if they are sending a log file to be viewed within the healthchecks webapp but thought it could be nice functionality should be fairly easy to implement.

boopzz:
I'd like to try and help out and add this in myself as it feels like it should be a fairly quick thing but cant for the life of me find where the values are set. All the vars are fairly common words in the code base like CODE, STATUS, TAGS 😅 . Any pointers to where the current ones are?

boopzz:
I think the functionality is there but not surfaced yet? https://github.com/healthchecks/healthchecks/blob/e5e369257c7e784ae4c0784c2f4390f0f3b57c6f/templates/integrations/webhook_form.html#L224

boopzz:
ok cool, might it be worth surfacing a checkbox on the form to say if you want it passed or not? I guess some people would want to send a bunch of logs to the UI but not forced into sending them via the notification

cuu508:
Potentially yes, but I'd say only if and when end users specifically request it. The Slack integration includes the last ping body with no opt-out option, and there have been no complaints.

PR #785, opened 2023-01-31
Amended Mattermost class to include the BODY in the message

boopzz:
PR for #784, very similar to #735 and #783
Now includes the BODY in the alert message

### DIFF
--- a/hc/api/tests/test_notify_mattermost.py
+++ b/hc/api/tests/test_notify_mattermost.py
@@ -59,3 +59,51 @@ class NotifyMattermostTestCase(BaseTestCase):
         self.channel.notify(self.check)
         self.channel.refresh_from_db()
         self.assertFalse(self.channel.disabled)
+
+    @override_settings(SITE_ROOT="http://testserver")
+    @patch("hc.api.transports.curl.request")
+    def test_it_shows_last_ping_body(self, mock_post):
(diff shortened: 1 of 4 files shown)


cuu508:
Thanks for the PR! 

Just tested this. There's an issue with the markdown formatting, looks like Mattrmost does not recognize \```pre-formatted-text\``` syntax in fields.

A notification with a message body in Slack:

image

Same notification in Mattermost:

image

But it looks like Mattermost *does* support pre-formatted text where each line is prefixed with 4 spaces:

image

But this syntax is not supported in Slack :grimacing: 

I think the Slack and Mattermost classes should each prepare the body differently: the Slack class should enclose it in backticks, the Mattermost class should prepend each line with 4 spaces.

Please also add test cases that exercise the new functionality.

cuu508:
Ah, there's a simpler fix–add newlines so that backticks are on separate lines.

Slack accepts either syntax:

    ```text```

and – 

    ```
    text
    ```

Mattermost only accepts the second variant. 

I'll merge this PR and tweak it a little bit.

boopzz:
ahh, I was going to say that MM has single backtick for inline and triple backtick for multiline, but you're quite right, on separate lines. You can add an interpreter to the first triple backtick and it will try and add syntax checking into the block but not needed in this case. Nice find 👍

PR #801, opened 2023-03-08
For email pings, "ping details" dialog box should default to show html if exists, otherwise to plain text.

seidnerj:
I believe the title says it all.


seidnerj:
Seems like this is failing due to network issues at github when trying to download dependencies...

seidnerj:
@cuu508, please re-run the checks later, I suspect this was a transient issue and the checks should succeed. Thanks!

cuu508:
Thanks for the PR, @seidnerj !

seidnerj:
My pleasure!

PR #817, opened 2023-04-04
Include body in msteams template if exists.

ch0wm3in:
Support it like it does in https://github.com/healthchecks/healthchecks/blob/master/templates/integrations/slack_message.json The markdown codeblock works well in teams also.


cuu508:
Thanks for the PR. 

I moved the body from the facts table to a separate section so it has more horizontal space.

PR #840, opened 2023-06-05
Fix smtpd service without stdin

GermanCoding:
Prior to 4b05fc6a8c5687640d30138a48a1403cc54898f4, the smtp daemon was working nicely as a daemon, i.e. without stdin attached (common in service scenarios like systemd). Since that commit, the daemon now relies on stdin being available and actually returning something that is **not** EOF, so you can't attach it to `/dev/null` or similar. You must have a full terminal available to run that as a daemon, which is not what a daemon usually is 😄 

This change removes the dependency on stdin and instead falls back on a more traditional SIGINT behaviour: Run until a SIGINT/CTRL-C is received, then exit.

### DIFF
--- a/hc/api/management/commands/smtpd.py
+++ b/hc/api/management/commands/smtpd.py
@@ -12,6 +12,8 @@ from django.db import connections
 from hc.api.models import Check
 from hc.lib.html import html2text
 
+import time
+
 RE_UUID = re.compile(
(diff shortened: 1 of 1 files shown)


cuu508:
Oops, I botched that one up :-/

Sorry for the trouble, and thanks for the speedy fix @GermanCoding !

GermanCoding:
Not a problem, thanks for the quick fix!

PR #845, opened 2023-06-16
Added fetch by slug to v3 api

scheibling:
Suggestion for fetch by slug using the ping key in the same way that it's done in the Ping API


cuu508:
Alternative implementation in https://github.com/healthchecks/healthchecks/commit/651aec4ac7ff240a20cc216d1df77a693fbdad0e

PR #854, opened 2023-07-02
Fix JS name and remove references to .map files

szepeviktor:
While doing typo fixes I've discovered `--exclude=static/js/*.min.js` misses zxcvbn.

Watch my next PR! 👁️


templates/accounts/set_password.html
```diff
@@ -44,7 +44,7 @@ <h1>Set a Password</h1>
 
 {% block scripts %}
 {% compress js %}
-<script src="{% static 'js/zxcvbn.js' %}"></script>
+<script src="{% static 'js/zxcvbn.min.js' %}"></script>
 <script src="{% static 'js/set-password.js' %}"></script>
 {% endcompress %}
```
szepeviktor, line 49: Should I move zxcvbn out of `compress js` ??

cuu508: No, `{% compress %}` bundles `zxcvbn` and `set-password` into a single file, which is what we want.

szepeviktor: Done 🙃

cuu508:
Thanks!

PR #872, opened 2023-08-03
Replace healtchecks with healthchecks

goldstar611:
Replace `healtchecks` with `healthchecks`


cuu508:
Thanks @goldstar611 !

PR #893, opened 2023-09-16
Switch to apt-get

solracsf:
More suitable and less error prone to use in scripts.
https://manpages.debian.org/buster/apt/apt.8#SCRIPT_USAGE_AND_DIFFERENCES_FROM_OTHER_APT_TOOLS


cuu508:
Thanks!

PR #899, opened 2023-09-24
Prevent auto hiding of the nav bar on Mobile Safari

seidnerj:
This can create situations in which certain buttons (e.g. ping body modal dialog "Got It!" button) appear in an area that is essentially un-clickable. When one clicks in that area, safari raises the nav bar and the underlaying button click does not happen (see attached screen recording as an example trying to click the "Got It!" button, top of the screen was cropped in order to remove potentially sensitive details)

https://github.com/healthchecks/healthchecks/assets/4147381/47ea3089-cc64-4a52-8bd7-3e7cce42ec87

I did check this commit on Safari (macOS 14/iOS 17.0.1), but am not sure how to test this thoroughly across multiple browsers and versions. Would appreciate some help/guidance on this.

### DIFF
--- a/static/css/base.css
+++ b/static/css/base.css
@@ -10,7 +10,7 @@ body {
 
 .footer {
     position: absolute;
-    bottom: 0;
+    bottom: auto;
     width: 100%;
     /* Set the fixed height of the footer here */
     height: 84px;
@@ -74,6 +74,18 @@ body {
 
 #global-links > li > a {
     font-size: small;
+}
+
+/* see here: https://stackoverflow.com/questions/23164503/disable-hiding-of-address-bar-on-mobile */
+@media only screen and (max-width: 500px) {
+
+    /* prevent auto hiding of the nav bar on mobile safari*/
+    body, html{
+        height: 100%;
+        position: fixed;
+        width: 100%;
+        overflow-y: scroll;
+    }
 }
 
 @media (min-width: 768px) {


cuu508:
Thanks for the PR!
I'd like to test this before merging but don't have an iPhone and not sure how to do that.

I do have a Mac with desktop Safari, but its responsive mode does not simulate the auto-hiding browser navbar.

seidnerj:
My pleasure. I think you might be able to check this via the iPhone simulator:

Screenshot 2023-09-25 at 12 47 04

cuu508:
Can reproduce the issue in simulator? It lets me scroll up and reach the "Got It" button

image

seidnerj:
bummer, you want to send me a public link in private and I'll try to check it on my end?

cuu508:
@seidnerj I understand you can reproduce the issue on a physical device. Can you also reproduce it in a simulator? If yes, it means I should be able to as well, and I'm just not doing something right.

seidnerj:
Checking now, I am definitely able to reproduce this on a physical device. Trying on a simulator in a bit.

seidnerj:
I was indeed able to reproduce this on a simulator, to be specific I used iPhone 12 Pro Max on iOS 17, see screen capture video attached. The key is to change safari layout to "Single Tab" (instructions here: https://support.apple.com/en-gb/guide/iphone/ipha9ffea1a3/ios).


https://github.com/healthchecks/healthchecks/assets/4147381/04611309-8101-4c87-9b7a-f6754c794c22

cuu508:
@seidnerj after changing Safari layout to "Single Tab" I can reproduce the issue in simulator!
But, when the navigation bar appears, I can still scroll the page up a little more and reach the button I intended to click.

seidnerj:
Yes, after it appears, you can indeed scroll up and click the button. The issue is, that this only happens after the user already clicked the button once and the click didn’t work. So the user ends up having to click-> scroll -> click which is a bad user experience. 

The simplest solution I found for this is to always show the nav bar on safari mobile so it always present. This way you can never reach the aforementioned scenario.

seidnerj:
any thoughts on this?

cuu508:
I understand the issue, but I'm not sure we should fix it. Yes, it's bad UX to have to click, scroll, and click again, but looks like that's how Apple intends it to work. You can reproduce the problem even on apple.com – if you scroll a page so that a clickable link is at the bottom of the screen, and try to click it, you will see the same behavior.

My issues with the proposed fix:

* it hacks browser's standard behavior. The navbar is supposed to hide, users expect the navbar auto-hiding on other pages, but here it is always visible, even when you are not trying to click anything near the bottom edge.
* the `@media only screen and (max-width: 500px) ` media query targets more than just mobile Safari users.

seidnerj:
I agree on both points. How about repositioning the button such that it is never in the bottom 50 or so pixels?

cuu508:
Good idea! I increased the bottom margin for modal windows on mobile screens from 10px to 60px. So now the user can:

1. scroll to the bottom
2. click on the button in modal's footer, and not trigger the nav bar

This affects all modal dialogs, not just the Ping Details one.

seidnerj:
Awesome! I’ll close this PR then. Thanks a lot!

PR #904, opened 2023-10-06
Try fixing mypy issues.

apollo13:
Refs https://github.com/typeddjango/django-stubs/issues/1755

### DIFF
--- a/hc/front/views.py
+++ b/hc/front/views.py
@@ -901,8 +901,8 @@ def details(request: AuthenticatedHttpRequest, code: UUID) -> HttpResponse:
         check.project.save()
 
     all_channels = check.project.channel_set.order_by("created")
-    regular_channels = []
-    group_channels = []
+    regular_channels: list[Channel] = []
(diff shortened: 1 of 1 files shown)


hc/api/admin.py
```diff
@@ -171,6 +171,10 @@ class ChannelAnnotations(TypedDict):
     owner_email: str
 
 
+def channels_view_on_site(obj: Channel) -> str:
+    return reverse("hc-channels", args=[obj.project_code])
```
apollo13, line 175: It complains here that it cannot find `project_code` which makes sense. But if I use WithAnnotations it also fails because then it doesn't match ModelT

apollo13: Okay and that fix is clearly wrong :)

cuu508:
One workaround idea I had was to define/override the `get_view_on_site_url` method on the admin model, instead of defining `view_on_site`. But it would be nice to get this fixed in django-stubs properly.

PR #930, opened 2023-12-16
Fix crash when email host not configured

### LINKED ISSUE #920: Disabling emails causes 500 error when adding new project members
Healthchecks should not raise 500 error when email configuration is not provided. It should not send any emails and fail gracefully.

```
Internal Server Error: /projects/142d4567-6d58-4484-ac27-31b447996fbc/settings/
Traceback (most recent call last):
  File "/lsiopy/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/lsiopy/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lsiopy/lib/python3.11/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapper_view
    return view_func(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/healthchecks/hc/accounts/views.py", line 428, in project
    if project.invite(user, role=invite_form.cleaned_data["role"]):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/healthchecks/hc/accounts/models.py", line 415, in invite
    profile.send_instant_login_link(membership=m, redirect_url=checks_url)
  File "/app/healthchecks/hc/accounts/models.py", line 146, in send_instant_login_link
    emails.login(self.user.email, ctx)
  File "/app/healthchecks/hc/lib/emails.py", line 84, in login
    send(make_message("login", to, ctx))
  File "/app/healthchecks/hc/lib/emails.py", line 67, in send
    assert settings.EMAIL_HOST, (
AssertionError: No SMTP configuration, see https://github.com/healthchecks/healthchecks#sending-emails
```

happens on any version >=v2.10

marlenekoh:
closes #920
Log an error instead of raising AssertionError when EMAIL_HOST is blank.

### DIFF
--- a/hc/accounts/models.py
+++ b/hc/accounts/models.py
@@ -411,8 +411,9 @@ class Project(models.Model):
         m = Member.objects.create(user=user, project=self, role=role)
         checks_url = reverse("hc-checks", args=[self.code])
 
-        profile = Profile.objects.for_user(user)
-        profile.send_instant_login_link(membership=m, redirect_url=checks_url)
+        if settings.EMAIL_HOST:
+            profile = Profile.objects.for_user(user)
+            profile.send_instant_login_link(membership=m, redirect_url=checks_url)
         return True
 
     def update_next_nag_dates(self) -> None:
--- a/hc/lib/emails.py
+++ b/hc/lib/emails.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import logging
 import time
 from email.utils import make_msgid
 from smtplib import SMTPDataError, SMTPServerDisconnected
@@ -10,8 +9,6 @@ from typing import Any
 from django.conf import settings
 from django.core.mail import EmailMultiAlternatives as Message
 from django.template.loader import render_to_string as render
-
-logger = logging.getLogger(__name__)
 
 
 class EmailThread(Thread):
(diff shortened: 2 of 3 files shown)


cuu508:
Thanks for the PR. With this change, instead of raising an exception, Healthchecks logs an error and skips sending email.

I would like to handle missing SMTP credentials higher up the call stack. In this instance, in `hc.accounts.models.Project.invite`. Currently:

``` python
    def invite(self, user: User, role: str) -> bool:
        [...]
        profile = Profile.objects.for_user(user)
        profile.send_instant_login_link(membership=m, redirect_url=checks_url)
        return True
```

After:

```python
    def invite(self, user: User, role: str) -> bool:
        [...]
        if settings.EMAIL_HOST:
            profile = Profile.objects.for_user(user)
            profile.send_instant_login_link(membership=m, redirect_url=checks_url)
        return True
```

Healthchecks sends different types of emails, and we may want to handle missing SMTP credentials differently for different types:

* invites: skip sending the email, and I don't think we should even log anything
* login links: we should not show the UI for requesting an invite email at all (but this can be a separate PR)
* email reports: `manage.py sendreports` should refuse to start if SMTP credentials are missing (but this can be a separate PR)
* ...

marlenekoh:
@cuu508 got it 🙂 to clarify, should I revert the behaviour of `hc.lib.emails.send()` to raise an `AssertionError` if `settings.EMAIL_HOST` is not configured?

cuu508:
@marlenekoh yes – the functions that call `hc.lib.emails.send()` (initially one, eventually all) will check SMTP credentials on their end. And so, if it does get called but has no SMTP credentials, that's an unexpected state and an exception is appropriate.

marlenekoh:
Done

cuu508:
Thanks!

PR #932, opened 2023-12-22
Updates to environment variables

mmomjian:
These changes quote true/false variables, as when unquoted they are susceptible to conversion to Booleans by YAML which is not compatible with BASH environment variables. [Source](https://docs.docker.com/compose/compose-file/compose-file-v2/)

It also adds an undocumented option which allows changing the internal port for the server. This is needed in certain VPN configurations where the port is not exposed using Docker and thus cannot be easily remapped from the default.


cuu508:
Thanks for the PR.

Regarding the boolean values, I'd like to understand this a little better. What would be a specific scenario where an unquoted value causes issues?

Regarding UWSGI_HTTP_SOCKET, I don't mind having it in .env.example, but I'd like to understand your use case here as well. What setup are you working with, and why remapping from the default is necessary?

Thanks!

mmomjian:
For the booleans, we can refer to YAML documentation sections [10.2.1.2](https://yaml.org/spec/1.2.2/#1021-tags). Non-quoted Boolean values in a YAML, such as true, false. yes, or no, are "bound to a native integer type, using one for true and zero for false." This means that the expected value of "True" is lost.

Since Docker variables can be specified as either an array or a dictionary, and this project uses the array synax (`- VAR=value`), I believe the boolean may not be typecast on all versions of docker compose. However for safety and to cover for people who prefer to specify their envvars as a (`dict: value`), I recommend quoting these potentially unsafe values.

For the socket, I run most of my docker containers attached to a wireguard container using the following compose entry:
```
    network_mode: "container:wireguard"
    depends_on:
      - wireguard
```
In this setting, using the `ports:` configuration in Docker Compose does not work, because that only changes the external port on the host, not the port within the docker network itself. When multiple services use the same port, like 8000 or 8080, this means that I cannot use them on the same Wireguard network for external access unless I change one of the internal ports.

cuu508:
> For the booleans, we can refer to YAML documentation sections [10.2.1.2](https://yaml.org/spec/1.2.2/#1021-tags). Non-quoted Boolean values in a YAML, such as true, false. yes, or no, are "bound to a native integer type, using one for true and zero for false." This means that the expected value of "True" is lost.

> Since Docker variables can be specified as either an array or a dictionary, and this project uses the array synax (- VAR=value), I believe the boolean may not be typecast on all versions of docker compose. However for safety and to cover for people who prefer to specify their envvars as a (dict: value), I recommend quoting these potentially unsafe values.

Sorry, I still don't quite understand. Are you saying Docker might in some situations convert unquoted `True` to `1` and unquoted `False` to `0`? Can you show a specific scenario where this would happen?

> For the socket, I run most of my docker containers attached to a wireguard container using the following compose entry [...]

Thanks for the explanation. uWSGI has many UWSGI_* env variables that we could potentially list here, but I don't think we should. The fact that you *can* configure uWSGI by setting env vars is mentioned in docs: https://healthchecks.io/docs/self_hosted_docker/ (section "uWSGI Configuration").

PR #944, opened 2024-01-21
Update filtering rules to specify that they are case sensitive

mmomjian:
We should specify that the keywords are case sensitive - this isn't obvious to the user. I determined this based on my own testing.

### DIFF
--- a/templates/docs/email.html-fragment
+++ b/templates/docs/email.html-fragment
@@ -5,7 +5,10 @@ sending email messages to special email addresses.</p>
 <p>By default, SITE_NAME will consider any email received at the displayed address as
 a "success" signal. You can also configure SITE_NAME to look for specific
 keywords in the subject line or the message body to decide if the message
-is a "start," a "success," or a "failure" signal. You can set up the keywords in
+is a "start," a "success," or a "failure" signal.</p>
+<p>Keywords are case-sensitive and filtered in the following order: SITE_NAME first
+looks for a failure, then success, then start. If filtering is enabled, an email
+containing no keywords will be ignored. You can set up keywords in
 the <strong>Filtering Rules</strong> dialog:</p>
 <p><img alt="Setting filtering rules" src="IMG_URL/filtering_rules.png" /></p>
 <h2>Use Case: Newsletter Delivery Monitoring</h2>
--- a/templates/docs/email.md
+++ b/templates/docs/email.md
@@ -8,7 +8,11 @@ sending email messages to special email addresses.
 By default, SITE_NAME will consider any email received at the displayed address as
 a "success" signal. You can also configure SITE_NAME to look for specific
 keywords in the subject line or the message body to decide if the message
-is a "start," a "success," or a "failure" signal. You can set up the keywords in
+is a "start," a "success," or a "failure" signal.
+
+Keywords are case-sensitive and filtered in the following order: SITE_NAME first
+looks for a failure, then success, then start. If filtering is enabled, an email
+containing no keywords will be ignored. You can set up keywords in
 the **Filtering Rules** dialog:
 
 ![Setting filtering rules](IMG_URL/filtering_rules.png)
--- a/templates/front/filtering_rules_modal.html
+++ b/templates/front/filtering_rules_modal.html
@@ -39,6 +39,7 @@
 
                 <div class="modal-body">
                     <h2>Inbound Emails</h2>
+                    <p>Matching is case-sensitive. <a href="{% url 'hc-serve-doc' 'email' %}">Lean more</a>.</p>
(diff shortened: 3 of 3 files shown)


mmomjian:
I also think a future PR could add a setting to ignore case. This can be achieved by adding the following code:

`check.no_match_case = form.cleaned_data["no_match_case"]` in hc/front/views.py:548

`def _match(subject: str, keywords: str, nomatchcase: bool) -> bool:` in hc/api/management/commands/smtpd.py:98 (and subsequent code in this function using .casefold() to ignore case)

Would also need to add a checkbox to the HTML template, and add the check.no_match_case when calling _match. I can throw together a PR if this would be desired.

cuu508:
Two comments:

1. Instead of this wording: 

> Comma-separated list of **case-sensitive** keywords. If subject or body contains any of the keywords, classify the email as "start".

I would prefer:

> Comma-separated list of keywords. If subject or body contains any of the keywords, classify the email as "start". **Matching is case-sensitive.**

2. With either variant, the text gets longer and now does not fit in two lines. To make it fit, we can make the column a little wider (change `col-sm-7` to `col-sm-8`) and make the label column a little narrower to make space (change `col-sm-4` to `col-sm-3`). Example (for demonstration purposes, applied to the first field only):

image

cuu508:
> I also think a future PR could add a setting to ignore case. 

Out of curiosity, in what specific scenarios would it be beneficial to ignore case?

cuu508:
> However now I am wondering if it makes sense to list the case matching at the top, under the checkboxes, so it is not repeated under each keyword?

Yes, that might be better. Would have to see how it looks. 

> We could also specify the precedence of sorting, which is currently failure > success > start. Without saying this, someone could be left wondering how a message containing both keywords will be treated.

Agreed, but we also have to be careful not to stuff too much into the form. The keyword matching precedence stuff could perhaps go into [docs](https://healthchecks.io/docs/configuring_checks/), and the form could have a "Learn more..." hyperlink pointing to it.


> I don't know all of the exact words the VM may use to indicate a failure with backing up, so my current failure keywords is: "fail,Fail,FAIL,error,Error,ERROR,down,Down,DOWN".

Thanks for explaining. Rather than guessing all possible failure keywords I think I would rather list the success keywords (if they appear in the email then I know for a fact the backup succeeded), and treat everything else as "not success". You do not need to catch all possible failure keywords. If you miss some, you will still get notified after the grace time passes.

An option to treat everything that is not success as an explicit failure is an interesting idea. The problem with that is, the more buttons and knobs we have, the harder it becomes for users to make sense of it all. I don't want to end up implementing mail filtering rules like dedicated mail clients have :-)

mmomjian:
I added a small <p> element under the header specifying the case sensitivity and placed a link to the emails section of the Docs. That's a good idea, if people need more detail that's a good place to put it.

I do understand the filtering rules and not wanting to make it more complex. Unfortunately for my use case I backup about 5 VMs per hypervisor. Each hypervisor has an email setup in HealthChecks. So if 4/5 of the backups succeed, HC will get 4 daily emails and one ignored email unless I catch the failure keyword. The gap between backups is not long enough for the grace time to catch it, since I don't think I can specify a different email for each VM backup.

cuu508:
Thanks for the PR, looks good!

> Unfortunately for my use case I backup about 5 VMs per hypervisor. Each hypervisor has an email setup in HealthChecks. So if 4/5 of the backups succeed, HC will get 4 daily emails and one ignored email unless I catch the failure keyword. The gap between backups is not long enough for the grace time to catch it, since I don't think I can specify a different email for each VM backup

Thanks for explaining. That's a tricky one indeed. I'll think about it, right now I don't have anything better to suggest than what you are doing already.

cuu508:
I'm experimenting with the UI in the "Filtering Rules" dialog. Here's the mockup I have currently. The note about case sensitivity is unfortunately gone for now at least–I want it to fit in a 1920x1080 screen without scrolling, and it is getting really tight.

image


@mmomjian if you have a minute, could you give the following questions a go? Would help me get an idea if I'm on the right track or if this is getting too complex.

* Looking at the screenshot, is it easy to understand what each control is intended to be used for?
* If a check was configured exactly like shown in the form, what would you assume would happen when a check receives a HTTP GET ping (so no request body)?
* What values would you set to get the following behaviour: consider all incoming pings failure signals unless the string "Backup successful" is present somewhere in the HTTP request body or email body

cuu508:
Thanks for checking it out, appreciate it!

Good idea about adding the "(case-sensitive)" bit.

There's not enough space on the form to describe the precedence rules when checking for success / failure keywords, but I tucked in a link to docs:

image

Issue #945, opened 2024-01-22
Grace Time not completely editable once a job is created


saschabrockel:
![image](https://github.com/healthchecks/healthchecks/assets/33376463/2e3418a7-c4a8-48a7-8ef8-5270e56c3b46)
Images can say more than words.

You can see that I want to change the grace time here but it's just possible to change it completely on creation. I cannot change the field from minutes to hours or anything else. It is not a button. It would be lovely to be able to edit this because recreation means another ID and changes in the script/app or whatever.

cuu508:
You can set it to 120 to get two hours, to 180 to get three hours, etc.

I will accept a PR which turns the minutes part into a dropdown, if:

* the end result looks good visually (both in light and dark mode)
* the code changes are not big, and it does not introduce new dependencies
* the dropdown works well with keyboard navigation

PR #953, opened 2024-02-06
Make Grace Time editable when job is created.

mickBoat00:
Fixes: #945


cuu508:
Thanks for the PR.

A few requests:

* It should be possible to make the input field and the combo box look like a single control, with no double border between them. The Bootstrap examples use a button element (https://getbootstrap.com/docs/3.3/components/#input-groups-buttons-dropdowns) but perhaps something similar is possible with `<select>`
* It's good to see the client side validation of the grace time, but the bounds are not quite right. The max value for grace time is 365 days.
* I would prefer to do the value*unit=grace_time_in_seconds calculation on the client side. Take a look at the "Add Check" dialog: 

[add-check-modal.js](https://github.com/healthchecks/healthchecks/blob/master/static/js/add-check-modal.js)

[add_check_modal.html](https://github.com/healthchecks/healthchecks/blob/master/templates/front/add_check_modal.html)

The `<select>` element uses multiplier values in the "value" attribute:

```html
<select id="add-check-period-unit" class="form-control period-input">
    <option value="60">minutes</option>
    <option value="3600">hours</option>
    <option value="86400" selected>days</option>
</select>
```

This makes conversion easy: multiply the input field's value with the select boxes value.

The `<select>` field also does not have a "name" attribute, and is not submitted to the server. Instead, every time its value changes, we recalculate the grace time in seconds and set it to a hidden field (which does get submitted to the server):

```js
    // Update the hidden field when user changes grace inputs
    $("#add-check-modal .grace-input").on("keyup change", function() {
        var secs = Math.round(grace.value * graceUnit.value);
        grace.setCustomValidity(secs <= 31536000 ? "" : "Must not exceed 365 days");

        if (secs >= 60) {
            $("#add-check-modal input[name=grace]").val(secs);
        }
    });
```

Perhaps the edit form can use the same technique?

mickBoat00:
Thank you for your comment, they are very useful. I will implement them right way.

mickBoat00:
@cuu508 please i have updated the code.Will be waiting for you feedback. 
Thank you very much.

cuu508:
The changes look great, nothing for me to nitpick on which is rare ;-)
Thank you!

Issue #853, opened 2023-06-29
Add badges for individual checks


kiler129:
It would be great to allow badges for individual checks. While there's a (documented) hack to assign a unique tag and get a badge for the tag, it doesn't really scale with hundreds of checks ;)

mickBoat00:
What happens when a tag and check name are the same? The same url for svg will be generated for instance.
Possible solution is to enforce uniqueness between check names and tags for a project. Is there a better solution?

mickBoat00:
What happens if a check name is the same as a tag for a different check in the same project. The same url will be generated for the svg for instance.Possible solution is solution is to enforce uniqueness between tags and different check names in a project. Is that good enough or there is a better solution?

PR #960, opened 2024-02-12
Checks badges

mickBoat00:


mickBoat00:
open to discussions about this solution i implemented.

cuu508:
cc:  #853

Thanks for the PR! It is awesome to see you taking on non-trivial tasks!

After looking around the changes, there are two improvements I'd like to suggest:

**1. Use different badge URL schemes for tag badges and check badges**

If we use check's name and tag's name in badge URLs we will run into ambiguous situations where a both a check and a tag with the same name exists. I saw you did add mitigation for this in the "Add Check" dialog, but we can still end up in this situation when editing existing checks, and when using Management APIs. Even if we manage to prevent the name overlaps everywhere, this would be a weird limitation to explain to the users.

For check-specific badges, we could use [Check.unique_key()](https://github.com/healthchecks/healthchecks/blob/b1cd5295323bce378259d316b4227d0d856c7fdc/hc/api/models.py#L372) values as identifiers. We already use them for [cloaked URLs](https://github.com/healthchecks/healthchecks/blob/b1cd5295323bce378259d316b4227d0d856c7fdc/hc/front/views.py#L989), and in [read-only API requests](https://github.com/healthchecks/healthchecks/blob/b1cd5295323bce378259d316b4227d0d856c7fdc/hc/api/views.py#L447).

The downside of using `unique_key` is that the database lookup is more expensive. `unique_key` is a calculated property, not a database field. We cannot easily fetch a check by its unique key, instead we have to grab all checks from the project, and iterate over them to find the right one. If this inefficiency starts to become a performance problem, we can move `unique_key` to a database field down the road.

Note – we cannot use check's UUID in the badge URL. If we did, anybody who has the badge URL would also be able to construct the check's ping URL. 

I don't have a concrete proposal for the badge URL scheme yet, I'll work on that.

**2. Redesign the "Badges" page**

When user loads the badges page, they see a badge for every tag. If we also add a badge for every check, the list of badges could get quite long. The page also makes three HTTP requests for each badge (for previewing SVG, JSON and Shields.io variants). If a project has 100 checks, the badges page would fire off 300+ HTTP requests.

I think we could redesign the page so that the user first selects which tag or check they want the badge for, then selects the badge format, and only then they see the badge preview and perhaps usage examples for embedding in HTML and .md documents.

A rough mockup to illustrate what I mean:

rect962

mickBoat00:
Thanks for the update i will look into it

cuu508:
The current URL scheme for a tag badge is:

```
badge/<badge_key>/<signature>/<tag>.<fmt>
```

And for all checks in the account:

```
badge/<badge_key>/<signature>.<fmt>
```

We cannot change that, we need to keep it working as is.

But I think  we can reuse the second one for check-specific badges too, and put check's unique_key in the signature field. In the badge view we can tell signatures and unique_keys apart by length. Signatures will be 8 or 10 characters, unique keys will be 40 or 40 characters.

A few examples:

```
# A 3-state SVG badge for the tag "emails"
https://healthchecks.io/badge/bd8e7a09-e0e7-4afb-8332-e2682e982a9d/JqyymSID/emails.svg

# A 2-state SVG badge for the tag "emails"
https://healthchecks.io/badge/bd8e7a09-e0e7-4afb-8332-e2682e982a9d/JqyymSID-2/emails.svg

# A 3-state SVG badge for all checks in the project
https://healthchecks.io/badge/bd8e7a09-e0e7-4afb-8332-e2682e982a9d/pezSzuu8.svg

# A 2-state SVG badge for all checks in the project
https://healthchecks.io/badge/bd8e7a09-e0e7-4afb-8332-e2682e982a9d/pezSzuu8-2.svg

# A 3-state SVG badge for one particular check:
https://healthchecks.io/badge/bd8e7a09-e0e7-4afb-8332-e2682e982a9d/124f983e0e3dcaeba921cfcef46efd084576e783.svg

# A 2-state SVG badge for all checks in the project
https://healthchecks.io/badge/bd8e7a09-e0e7-4afb-8332-e2682e982a9d/124f983e0e3dcaeba921cfcef46efd084576e783-2.svg
```

Another thought, this could perhaps be organized in two PRs:

* A PR that adds support for generating check-specific badges (but does expose the new URLs in the "Badges" screen yet)
* A PR that redesigns the "Badges" screen

cuu508:
@mickBoat00 do you plan to do more work in this? If not, I'll jump on it. Since playing with the the mockup I'm itching to get it implemented ;-)

mickBoat00:
> @mickBoat00 do you plan to do more work in this? If not, I'll jump on it. Since playing with the the mockup I'm itching to get it implemented ;-)

please go ahead

Issue #957, opened 2024-02-10
Auto-refresh for Log page


asdffdsazqqq:
hello and thanks for a great project.
currently, `https://healthchecks.io/checks/uuid/details` will auto-refresh the display as new checks come inbound.
however   `https://healthchecks.io/checks/0ea5c6f3-f103-4f85-9d6e-c40524578241/log/` will not auto-refresh.

so please, enable auto-refresh to log page.

thanks much, david

cuu508:
@asdffdsazqqq implementing auto-refresh is doable, but [gets quite tricky complexity-wise](https://github.com/healthchecks/healthchecks/pull/963#issuecomment-1951139277) if we want to do it efficiently. 

We can also consider a smaller, simpler feature: after the log page is loaded, periodically check if there are any new events. If there are, instead of fetching and displaying them, show a message near the top of the page that says something like "There are new log events, reload the page to see them". This would be not as fancy, but much easier to implement. 

What are your thoughts, would the simpler feature be useful? Could it substitute full auto-refresh for you?

PR #963, opened 2024-02-17
Add auto-refresh functionality to log page

mickBoat00:
Fixes: #957

### DIFF
--- a/hc/front/forms.py
+++ b/hc/front/forms.py
@@ -384,10 +384,8 @@ class SearchForm(forms.Form):
 
 class SeekForm(forms.Form):
     # min_value is 2010-01-01, max_value is 2030-01-01
-    start = forms.IntegerField(
-        min_value=1262296800, max_value=1893448800, required=False
-    )
-    end = forms.IntegerField(min_value=1262296800, max_value=1893448800, required=False)
+    start = forms.FloatField(min_value=1262296800, max_value=1893448800, required=False)
+    end = forms.FloatField(min_value=1262296800, max_value=1893448800, required=False)
 
     def clean_start(self) -> datetime | None:
         if self.cleaned_data["start"]:
--- a/hc/front/forms.py
+++ b/hc/front/forms.py
@@ -384,14 +384,20 @@ class SearchForm(forms.Form):
 
 class SeekForm(forms.Form):
     # min_value is 2010-01-01, max_value is 2030-01-01
-    start = forms.IntegerField(min_value=1262296800, max_value=1893448800)
-    end = forms.IntegerField(min_value=1262296800, max_value=1893448800)
+    start = forms.IntegerField(
+        min_value=1262296800, max_value=1893448800, required=False
+    )
+    end = forms.IntegerField(min_value=1262296800, max_value=1893448800, required=False)
 
     def clean_start(self) -> datetime:
-        return datetime.fromtimestamp(self.cleaned_data["start"], tz=timezone.utc)
+        if self.cleaned_data["start"]:
+            return datetime.fromtimestamp(self.cleaned_data["start"], tz=timezone.utc)
+        return None
 
     def clean_end(self) -> datetime:
-        return datetime.fromtimestamp(self.cleaned_data["end"], tz=timezone.utc)
+        if self.cleaned_data["end"]:
+            return datetime.fromtimestamp(self.cleaned_data["end"], tz=timezone.utc)
+        return None
 
 
 class TransferForm(forms.Form):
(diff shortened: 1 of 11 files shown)


cuu508:
Thanks for the PR.

IIUC, this auto-refresh implementation does almost all the same work as reloading the page would do:

* It fetches the full list of ping records (up to 1000 records)
* It reinitializes the slider
* It re-registers the onclick handler for individual rows
* It runs the timezone conversion for all rows

To save both the server and client resources, I'd like the auto-refresh to:

* Only fetch rows that have appeared since the last request
* Reinitialize the slider only when the current wall clock minute changes (so ~ once per minute)
* Use event delegation for the click handler, so the click handler can be registered once on page load
* Run the timezone conversion only on the new rows, not the whole table

I understand this gets quite complex. Perhaps we are nearing a point where we need to look at state management libraries, not sure.

We can also consider a smaller, simpler feature: when the log page is loaded, periodically check if there are any new events. If there are, instead of fetching and displaying them, show a message near the top of the page that says something like "There are new log events, reload the page to see them". This would be not as fancy, but *much* easier to implement.

mickBoat00:
Thanks for your valuable feedback, i gain a lot from it. I think it will be possible to be handle event delegation, fetches and update just a few rows and reinitialize the slider per minute without a state managements library. I implement these suggestion to see the improvements, they bring.

cuu508:
For event delegation, the feature I mean is adding an event listener to a parent element, so the parent element can receive events that bubble up from dynamically added child elements. For jQuery, this is done via the `selector` argument when adding a listener: https://api.jquery.com/on/

An example in Healthchecks codebase where we already do this: https://github.com/healthchecks/healthchecks/blob/52a8df4dbf60c73959e12473d67626391a0d92f7/static/js/details.js#L132

(we add an event listener to the `#events` element but look for clicks on `<tr class="ok">` child elements)

mickBoat00:
Hi @cuu508, I have a quick question. Before that, I've updated the code – not the final version, but it currently employs event delegation to handle click events for the child elements, fetches a few event rows, performs timezone conversion for those rows, and finally updates the slider every minute. Feel free to comment on the code changes.

My first question is: what should happen when the slider is moved and a new log arrives? I was thinking that we could display a message on the page saying, "New log events have arrived. Please reload the page to view them." Resetting the filtering that the user is performing by moving the slider every time new logs arrive may not provide a good user experience. What are your thoughts on this?

My second question pertains to the scenario where a check has no events yet, so there is no "see more" link to the log page. However, if the user manually appends "/log" to the browser URL to go to that page, is it okay to have Table elements in the DOM even though there are no events yet? This would make it easier to insert event rows when new logs arrive.

cuu508:
> My first question is: what should happen when the slider is moved and a new log arrives?

I think we should distinguish between the case where the user has moved the "end" slider handle, and where they have not.

If the user has moved the slider handle, we should stick to their selection, and perform no auto-refresh (we are looking at events that happened in the past, no new events can pop up there).

If the user has *not* moved the handle, I think we should treat its default value as "current wall clock time" and adjust it forward during auto-refresh.

To make this logic obvious to the user, perhaps we should even have a "Live updates: 🟢" / "Live updates: 🔴" indicator in the UI. It would be green by default, but would switch to red as soon as the user moves the end slider handle.

Currently, when user moves any handle, we add `start` and `end` query parameters to the URL. We could perhaps instead add just the query parameter for the one handle that gets moved. Then we could check if `end` is in the query string to determine if the user has moved the end handle or not.

> is it okay to have Table elements in the DOM even though there are no events yet?

Yes, that's fine, it's a tiny amount of markup.

mickBoat00:
Hi @cuu508, I've made some changes to the code. This time, I'm using session storage to maintain the state of the slider movement. As we all know, the log page can be tricky because it refreshes when a slider is used, so using a variable doesn't work well.

Additionally, live updates stop when either the start or end handle is moved. I understand that the ultimate goal is to have live updates even when the start handle is moved but not the end handle. This means we'll need to prevent the end timestamp from being added as request parameters to the view by disabling the form input field.

I'd like to hear your thoughts on the current implementation and whether this is a step in the right direction.

mickBoat00:
Thanks, @cuu508! Even though most of my code underwent changes, it was awesome working on this feature. I learned a lot. I'll look into the changes to see how to approach my next issues.

kristof-mattei:
@cuu508 I'd love to learn about the switch from integer to float for the datetime. What kind of issues did you run into?

cuu508:
@kristof-mattei It is for more precision when fetching live updates.

The Log page regularly polls (~ every 3s) the server for new pings or notifications. It passes a timestamp to the server ("give me all new events *after* this timestamp). If the timestamp has seconds precision, we could either miss updates, or load same updates twice when multiple events happen in the same second.

Increasing the timestamp precision is a simple fix. An alternative solution would be instead of passing a timestamp to pass the id of the most recent event we already have. This would get a bit hairy though as we are showing two types of events – pings and notifications, and they have different sequences of ids.

Issue #873, opened 2023-08-05
Filter events of a check


BeyondVertical:
It would be great to filter the events of a check (in the "Show More" page in the Events section).
So we could filter by success or log or error state and find faster the error loggings or the log events.

PR #968, opened 2024-03-04
Add events filter feature to log page.

mickBoat00:
Fixes: #873

### DIFF
--- a/hc/front/views.py
+++ b/hc/front/views.py
@@ -208,6 +208,60 @@ def _get_referer_qs(request: HttpRequest) -> str:
         return "?" + parsed.query
     return ""
 
+def _get_events_filter(events: list[Notification | Ping]) -> list[tuple[str, str]]:
+    events_counts: dict[str, int] = defaultdict(int)
+    output = []
+    for event in events:
+        if isinstance(event, Notification):
+            events_counts["alerts"] += 1
+        else:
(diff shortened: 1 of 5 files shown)


mickBoat00:
@cuu508 what is your thoughts on this implementation

cuu508:
Hello @mickBoat00, thank you for the PR!

Healthchecks is a mature project, and there are few "easy pickings" in the issue tracker. Most open feature suggestions fall into one of the two categories:

* I'm not sure I want the feature in the project at all. Maybe sometime in the future, so I'm not closing it.
* The feature would be nice to have, but would take effort to design and implement. To implement it well, other parts of the system may need to be rethought and redesigned as well. Programming is hard so I rather go shopping :-)

This one falls in the second category. Problems with the proposed implementation:

* The UI becomes more cluttered, and it is not obvious that the badges in the "Filter events by:" line can be clicked
* Does not play well with auto-refresh. 
* There's a bunch of new code. Can it be simplified and reduced to, say, 20 lines of new code?

On the UI front, one idea I'm thinking is to drop the time slider, and replace it with a couple date pickers. These could go in a "Filter" side panel, similar to filters in Django Admin. The side panel then would be a logical place to put the event type filters too. A quick mockup:

image

mickBoat00:
Thanks for your feedback

Issue #967, opened 2024-03-03
Unable to send email notification with protonmail bridge


nmonserrato:
Hey there!
I'm running healthchecks on a TrueNas Scale instance (docker image / k3s), and trying to configure it so that it uses my local protonmail bridge to send the email notifications.

I'm quite sure that the bridge works fine because I also use it for Truenas alerts, and they work ok.

The bridge unfortunately uses a [self-signed certificate](https://proton.me/support/protonmail-bridge-clients-windows-thunderbird), and from the pod logs it seems clear that this is the issue:

```
2024-03-03 20:58:57.737783+01:00Exception in thread Thread-1:
2024-03-03 20:58:57.737833+01:00Traceback (most recent call last):
2024-03-03 20:58:57.737842+01:00File "/usr/lib/python3.11/threading.py", line 1045, in _bootstrap_inner
2024-03-03 20:58:57.738401+01:00self.run()
2024-03-03 20:58:57.738427+01:00File "/app/healthchecks/hc/lib/emails.py", line 26, in run
2024-03-03 20:58:57.738526+01:00self.message.send()
2024-03-03 20:58:57.738536+01:00File "/lsiopy/lib/python3.11/site-packages/django/core/mail/message.py", line 300, in send
2024-03-03 20:58:57.738684+01:00return self.get_connection(fail_silently).send_messages([self])
2024-03-03 20:58:57.738728+01:00^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-03-03 20:58:57.738737+01:00File "/lsiopy/lib/python3.11/site-packages/django/core/mail/backends/smtp.py", line 128, in send_messages
2024-03-03 20:58:57.738834+01:00new_conn_created = self.open()
2024-03-03 20:58:57.738867+01:00^^^^^^^^^^^
2024-03-03 20:58:57.738884+01:00File "/lsiopy/lib/python3.11/site-packages/django/core/mail/backends/smtp.py", line 93, in open
2024-03-03 20:58:57.738954+01:00self.connection.starttls(context=self.ssl_context)
2024-03-03 20:58:57.739000+01:00File "/usr/lib/python3.11/smtplib.py", line 790, in starttls
2024-03-03 20:58:57.739257+01:00self.sock = context.wrap_socket(self.sock,
2024-03-03 20:58:57.739280+01:00^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #972, opened 2024-03-11
feat: env var to ignore smtp server certificates 

nmonserrato:
https://github.com/healthchecks/healthchecks/issues/967

Added the possibility to skip server certificate validation with an env var (default False).

Tested running the server locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,12 @@ environment variables:
     EMAIL_USE_TLS = True
     ```
 
+If your SMTP server uses a self-signed certificate (e.g. protonmail-bridge), you can instruct healthchecks to ignore invalid certificates by setting:
+
+  ```python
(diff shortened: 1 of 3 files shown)


cuu508:
Thanks for the PR, but I don't think this functionality should live in Healthchecks. It deals with Django's email sending functionality, and is not Healthchecks-specific. Any Django project could face a similar situation. I think it should either live in the Django project (if the Django project accepts it) or in a Django app.

dl-sdu:
@CodiumAI-Agent /review

dl-sdu:
@CodiumAI-Agent /improve

cuu508:
@dl-sdu what is going on, what are you doing?

spinozzo:
> #967
> 
> Added the possibility to skip server certificate validation with an env var (default False).
> 
> Tested running the server locally.

very nice

PR #977, opened 2024-03-21
Proposing addition of a visual indication to favicon on down

princekhunt:
Hello,

I have tried adding a visual indication to favicon.
When monitor goes down, A new favicon with the count (number of down monitors.) gets updated.

Screenshot


cuu508:
Thanks for the PR.

1. Stylistically, it would be nice to superimpose the indicator over the existing favicon
2.  One another place where we dynamically change the document title is the individual check's Details page: it is either "{check_name} – {site_name}" or "DOWN – {check_name} – {site_name}". It would be nice to update favicon there as well.

With the above two in mind, I would prefer to go for a simpler approach and show an exclamation mark or a bell icon or just a red dot in the favicon instead of a number. 

Instead of drawing the favicon in canvas we could then have two png/ico/svg files in the static files directory, and switch between them in JS.

cuu508:
I think we could switch to SVG favicons as well:

favicon
favicon-warning

cuu508:
A simpler red-dot version:

favicon-warning-2

princekhunt:
Hello, Thank you for the review.

1. I have added 2 svg favicons, replaced the ico favicon with the svg favicon. 
2. Added the favicon updation in details.js as well.

Kindly review.


Thanks

Issue #939, opened 2024-01-08
Truncate beginning of long request bodies instead of the end


StevenMassaro:
I've started making use of the [feature to send log data to healthchecks](https://healthchecks.io/docs/attaching_logs/). I like that I get a truncated copy of the logs in my notification provider.

Under most circumstances I can imagine, an error in the script would show up at the end, and typically this error is truncated away.

I'm proposing a change to the truncating logic [here](https://github.com/healthchecks/healthchecks/blob/master/hc/api/transports.py#L70), something like this:

```
    if body and maxlen and len(body) > maxlen:
        body = body[-maxlen:] + "\n[truncated]"
```

but I wanted to get feedback first before spending time on this change.

This would ensure that the end of the log message is included in the notification, and the beginning of the log message is omitted.

cuu508:
Hi @StevenMassaro, thanks for the suggestion.

We truncate the request bodies for two main reasons:

1. to prevent storage from ballooning
2. to make the handling  of each individual ping request quick and efficient

For (1) it would not matter if we keep the head or the tail part of the request body. 
For (2) it does matter: suppose the client sends 1GB request body, we process it all and keep the last 100kB. This would be a waste of time, bandwidth and processing power, and would complicate the code (loading the entire request body in memory would be a bad idea, so `body = body[-maxlen:]` would not work). This is an extreme example but even sending 10MB and keeping last 100kB would not be great. 

I suggest doing one of the following:

- make your log output less verbose so it fits in 100kB
- do the truncation on the client side. [runitor](https://github.com/bdd/runitor), for example, does this out of the box: it truncates the log output on the client side, and sends the *tail* not the the head part to the server.

StevenMassaro:
Hi @cuu508 

I think I must be missing something with regards to your #2.

As the code currently reads to me, it looks like the body is loaded into memory on line 67, then truncated later. Thus, it shouldn't matter whether we retain the head or tail of the body, since it is already loaded into memory.

Let me also clarify my goal here. I am already doing truncation on the client side, as described in the documentation. My problem is that the request body is further truncated when a message is sent via telegram ([see code here](https://github.com/healthchecks/healthchecks/blob/master/hc/api/transports.py#L455)), and it is omitting the error message, because it is retaining the head, not the tail of the message:

![image](https://github.com/healthchecks/healthchecks/assets/25826807/933fdaf8-f9aa-41be-931f-8c840b212ec7)

cuu508:
[...]

PR #988, opened 2024-04-11
Add support for custom notification truncate settings

moraj-turing:
Hello, this PR proposes two basic settings added to a check configuration as a flexible solution to #939 

image


cuu508:
Thanks for the PR, but I think the right approach here is to make a decision what to truncate, and not make it user-configurable.

Downsides to making it user-configurable:

* this is a niche option, few users will use it. But it will clutter the UI for all users.
* makes product harder to understand. The UI mockup above does not explain when the truncation happens (during ping ingestion, when sending notifications, some other time?) 

Also note that the "truncate logs after" limit is dictated by the maximum payload size we can send to various integration channels. Telegram, for example, has 4096 character message limit, other integration types have different limits. Allowing the user to set an arbitrary limit or no limit would result in users sometimes shooting themselves in the foot.

For ping bodies included in notifications, I'm leaning towards truncating the beginning (and keeping the end) like @StevenMassaro in #939 suggested.

PR #989, opened 2024-04-11
Disable magic link login if SMTP host is not configured

moraj-turing:
This is a simple one! Added a basic condition in the django template to check if `EMAIL_HOST` is set to render the magic link form.


hc/accounts/views.py
```diff
@@ -208,6 +208,7 @@ def login(request: HttpRequest) -> HttpResponse:
         "registration_open": settings.REGISTRATION_OPEN,
         "support_email": settings.SUPPORT_EMAIL,
         "account_closed": "account-closed" in request.GET,
+        "magic_link_enabled": settings.EMAIL_HOST not in (None, ""),
```
cuu508, line 211: My naming preference, used in various places in this project, is `use_<somefeature>`. In this case, I suggest `"use_magic_form": ...`.

`settings.EMAIL_HOST not in (None, "")` will miss another falsy value, `False`. I think we should test for a falsy value like this instead: `bool(settings.EMAIL_HOST)`.

templates/accounts/login.html
```diff
@@ -29,7 +29,8 @@ <h1>Sign In to {{ site_name }}</h1>
         </div>
         {% endif %}
 
-        <div class="row">
+        <div class="row" style="display: flex; justify-content: center;">
```
cuu508, line 32: Please move the inline CSS to a CSS file.

cuu508 (changes requested):
Thanks for the PR! There are a couple small things I'd like to change.

templates/accounts/login.html
moraj-turing: So given its just two rules, I tried implementing this with bootstrap classes but bootstrap has flexbox support by default until v4+. I tried a drop-in replacement but that didn't help either.

moraj-turing:
Changes done!

cuu508:
Thanks for making the changes. I just tested them locally and noticed another thing that needs to be addressed: if the viewport is narrow (say, phone screen), the two login forms do not stack vertically any more (I think due to the addition of the `display: flex`).

I think the way to work around this is to remove the CSS changes, and center the form using [bootstrap's offset classes](https://getbootstrap.com/docs/3.3/css/#grid-offsetting). If EMAIL_HOST is not set then the magic login form will be absent. To make sure the email/password form is centered, we can do something like:

```
<div class="col-sm-6{% if not use_magic_form%} col-sm-offset-3{% endif %}">
```

cuu508:
Fixed in https://github.com/healthchecks/healthchecks/commit/1bdfbac775f93783f7500503ad9662faf0dfe7ca

PR #992, opened 2024-04-14
docs: add gitlab.com/etke.cc/go/healthchecks example to go page

etkecc:
Battle-tested in prod for etke.cc services :)


cuu508:
Thanks for the PR! (and for your work on the library!)

I would prefer this to go to the resources page in docs: https://github.com/healthchecks/healthchecks/blob/master/templates/docs/resources.md Could you add it there instead?

cuu508:
Thanks!

PR #1011, opened 2024-06-07
Addition of "True Read Only" Role

### LINKED ISSUE #1005: Read only user can create project, can we have "true" read only users ?
Hi, 

First : Thanks for this amazing software !

Everything is in the title, When I share a project by mail with someone, he can create is own projects and checks even with the fact that registration are disabled. I would like to have to possibility of "true" read only user who can only see the checks and do nothing.

tgrl84:
Fixes #1005
We are geco-IT, 
This pull request add a role called "True Read Only". 
If a user is assigned to this role in a project, they are prevented from creating new projects, even if they have other roles in different projects.
Capture d’écran du 2024-06-07 14-22-36

### DIFF
--- a/hc/accounts/models.py
+++ b/hc/accounts/models.py
@@ -493,7 +493,6 @@ class Member(models.Model):
     def is_true_rw(self) -> bool:
         return self.role in (Member.Role.TRUEREADONLY)
 
-
 class Credential(models.Model):
     code = models.UUIDField(default=uuid.uuid4, unique=True)
     name = models.CharField(max_length=100)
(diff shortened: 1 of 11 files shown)


cuu508:
I don't think having another project membership role is the way to go here. 

I think "Can create projects" should be a permission that belongs to the User object or to the Profile object, not to a particular membership. I have not thought in-depth about this but two initial ideas I have are:

* have a site-wide configuration switch which allows or forbids project creation for regular users (but does not apply to superuser, superuser)
* have a configurable default check limit for new users. On Healthchecks.io it is currently hardcoded to 20, and on self-hosted instances it is currently [hardcoded to 10000](https://github.com/healthchecks/healthchecks/blob/24e5e83bbcdd221e99642c0265251f8cbbf0bf0d/hc/accounts/models.py#L61). Site operator could set it to 0, and manually bump it up to any accounts that do need the ability to create checks.

If you are planning to work on this, and want to see your changes merged in this repository,  please let's discuss the plan first.

PR #1013, opened 2024-06-10
Add SHOW_DOC Environment Variable

tgrl84:
This pull request introduces a new environment variable, SHOW_DOC. 
This variable allows to control the display of documentation links.
This provides greater flexibility in configuring the user interface based on the needs of different deployment environments.
Capture d’écran du 2024-06-10 10-55-33


cuu508:
Can you share details about your use case – why would you prefer not to have the documentation section, or the footer in your deployment?

tgrl84:
We need to remove the documentation because we're sharing the project with users who don't need access to it. This is related to the true read-only role. Similarly, we've removed the footer.

Issue #1008, opened 2024-06-03
check display


pmatuszy:
I have very old logs (even from 2022). In the replication log I can see only month and day
![image](https://github.com/healthchecks/healthchecks/assets/58336955/ccc505c2-3207-4893-9a44-2ae3cd5f7b40)

It suggest that all checks from the current year should be in format Mon DD (e.g. Jun 01) but from the previous years it should include in the format also year (Mon DD, YYYY) 
or
just to remove ambiguity and display a year next to all logs.

PR #1015, opened 2024-06-11
format dates in logs to include year unless they are from current year

NKETIAH53:
fixes: #1008

### DIFF
--- a/static/js/details.js
+++ b/static/js/details.js
@@ -141,14 +141,19 @@ $(function () {
         return false;
     });
 
+    var current_year = moment().year() 
     function switchDateFormat(format) {
         lastFormat = format;
 
         document.querySelectorAll("#log tr").forEach(function(row) {
             var dt = moment.unix(row.dataset.dt).utc();
             format == "local" ? dt.local() : dt.tz(format);
+            var dateFormat = "MMM D"
+            if (current_year != dt.year()) {
+                dateFormat = "MMM D, YYYY"
+            }
 
-            row.children[1].textContent = dt.format("MMM D");
+            row.children[1].textContent = dt.format(dateFormat);
             row.children[2].textContent = dt.format("HH:mm");
         })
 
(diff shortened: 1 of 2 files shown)


static/js/details.js
```diff
@@ -147,8 +147,12 @@ $(function () {
         document.querySelectorAll("#log tr").forEach(function(row) {
             var dt = moment.unix(row.dataset.dt).utc();
             format == "local" ? dt.local() : dt.tz(format);
+            var dateFormat = "MMM D"
+            if (moment().year() != dt.format("YYYY")) {
+                dateFormat = "MMM D YYYY"
```
cuu508, line 152: IMO the date looks better with a comma separator before year – "MMM D, YYYY", could you add it?

static/js/details.js
```diff
@@ -147,8 +147,12 @@ $(function () {
         document.querySelectorAll("#log tr").forEach(function(row) {
             var dt = moment.unix(row.dataset.dt).utc();
             format == "local" ? dt.local() : dt.tz(format);
+            var dateFormat = "MMM D"
+            if (moment().year() != dt.format("YYYY")) {
```
cuu508, line 151: * Move the `moment().year()` part outside the function, so it is not evaluated again for every row
* Instead of `dt.format("YYYY")`, do `dt.year()`

cuu508 (changes requested):

static/js/details.js
NKETIAH53: Sure, will do that

static/js/details.js
NKETIAH53: Noted

cuu508:
Thanks!

dl-sdu:
@CodiumAI-Agent /review /improve

dl-sdu:
@CodiumAI-Agent /improve

Issue #1020, opened 2024-06-28
No warning or info on setting "APPRISE_ENABLED=True" and apprise not installed


aslirajesh:
**Issue** 
When user set APPRISE_ENABLED=True and "appraise" is not installed in venv, there is no message to install "apprise" to use. It silently disabling the apprise integration without any message.

**Fix**
Either add "apprise" in requirements.txt or show proper message on starting to install apprise when APPRISE_ENABLED is True .

**Code changing APPRISE_ENABLED False without warning**
[hc/api/tranports.py](https://github.com/healthchecks/healthchecks/blob/8d0930c4b9a0fd4124f1c15b93dc4e9bd0a6ab62/hc/api/transports.py#L36-#L40)

PR #1021, opened 2024-06-29
Show warning if apprise is enabled but apprise package is not installed

aslirajesh:
## Description
Added  warning message if APPRISE_ENABLED is set to 'True' in settings, and python package 'apprise' is not installed.

Before changes APPRISE_ENABLED was being set to false without warning when apprise was not installed, even if it is enabled in settings.py

## Issue 
Fixes: #1020 

## Checklist
- [ ] Issue Discussed
- [X] self-review
- [X] Format check of changed files with black

## Screenshot

Screenshot 2024-06-29 122231

### DIFF
--- a/hc/api/apps.py
+++ b/hc/api/apps.py
@@ -43,6 +43,29 @@ def settings_check(
 
 
 @register()
+def settings_check(
+    app_configs: Sequence[AppConfig] | None,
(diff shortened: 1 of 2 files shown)


cuu508:
Thanks @aslirajesh !

PR #1034, opened 2024-07-30
Fix django version in self hosted documentation

jpizquierdo:
Fix django version in self hosted documentation From 4 to 5


cuu508:
Good catch, thanks!

PR #1035, opened 2024-07-30
Improvements for self-host deployment

Kvark1205:


cuu508:
Thanks for the PR. A few requests:

* Please submit changes to .env file, to Dockerfile and to docker-compose file as separate PRs.
* I would prefer to keep entries in .env file sorted alphabetically
* I noticed you changed `ALLOWED_HOSTS=localhost` to `ALLOWED_HOSTS=*`. This is bad for security and should not be the default.
* In PR comments please describe precisely what the changes are intended to achieve. For example, what is the upside of adding `--no-cache-dir`? Is the resulting image smaller? By how much?
* A change in docker-compose switches postgres image from `postgres:12` to `postgres:15.7-alpine` and changes the container name. With this change, what will be the consequences for end users? When they upgrade, will their existing postgres databases get seamlessly upgraded? Have you tested this?

Thanks!

cuu508:
No reply, closing.

Issue #1054, opened 2024-09-02
ERROR: Too many open files


bilalvadion:
Healthcheck ping returned a 500 internal server error to my client app. Upon investigation, I found the following error:

`OSError: [Errno 24] Too many open files:`

Please find syslog subset in log file attached below. 
[too-many-open-files.log](https://github.com/user-attachments/files/16834389/too-many-open-files.log)

I have tried increases the ulimit, but issue still persists.
Reference: https://stackoverflow.com/questions/39537731/errno-24-too-many-open-files-but-i-am-not-opening-files

Current ulimit is 65535.

Environment:
Ubuntu 20.04.6 LTS
Python 3.8

cuu508:
Can you please check if the limit is in fact 65535:

```
cat /proc/<pid>/limits | grep files
```

Also, please check what files are in fact open:

```
lsof -p <PID>
```

bilalvadion:
@cuu508 here are the command results:

```
lsof -p 3887889 | wc -l
78
```

```
cat /proc/<PID>/limits | grep files
Max open files            1024                 524288               files     
```

```
cat /proc/<PID>/limits | grep files
Max open files            1024                 524288               files     
```

```
ulimit -Sn
65535
```

```
ulimit -Hn
65535
```

```
ulimit
unlimited
```

```
ulimit -n
65535
```

cuu508:
`/proc/<PID>/limits` shows the soft limit is 1024 and the hard limit is 524288. Neither is 65535, so something's not right with your ulimit usage.

Can you reproduce the "Too many open files" error? If yes, please run `lsof -p <pid>` then send me the full output (you can send it privately to contact at healthchecks io). I'd like to figure out how the file handles are being spent.

bilalvadion:
I increased the soft limits using the steps below:

```
sudo nano /etc/systemd/system.conf

#DefaultLimitNOFILE=1024:524288
to
DefaultLimitNOFILE=65535:524288

sudo systemctl daemon-reload //to reload config
systemctl restart healthchecks-runserver.service //to restart service
ps aux | grep python //to check process id
cat /proc/<PID>/limits | grep files //to check if limit increased form 1024 to 65535
```

But I had an unexpected sideaffect on the healthchecks service, below error keeps repeating per ping request:

`Sep  4 00:00:00 ip-172-31-6-192 python3.8[3902760]: ----------------------------------------
[...]

PR #1056, opened 2024-09-05
Minimal production setup using uWSGI

BilalKamran:


BilalKamran:
Resolution of issue: https://github.com/healthchecks/healthchecks/issues/1054

Issue #1074, opened 2024-10-20
users passed from REMOTE_USER_HEADER are case sensitive, which conflicts with other functions


rophy:
Version: 3.5.1

Issue: healthchecks normalizes email addresses to lower cases in most of the place, but missed to do that for REMOTE_USER_HEADER. Thus, when setting up oauth2-proxy with healtchecks, using an OIDC provider which populates emails in upper cases, conflicts will happen.

Example:
1. Authenticated users get account name USER01@GMAIL.COM
2. USER01 invites USER02@GMAIL.COM to project -> when saving, the invite got converted to user02@gmail.com
3. USER02 logs in as USER02@GMAIL.COM
4. Result: USER02 cannot see USER01 project

cuu508:
Thanks for the report and the PR.

I'm worried this could break existing setups. Example:

Alice is the only user of a Healthchecks instance, their address is USER01@GMAIL.COM and they have existing checks and integrations already set up.

With this fix applied, 

* Alice visits their Healthchecks instance
* Their email address gets normalized to user01@gmail.com
* This address does not exist in the users table, so Healthchecks creates a new user account and logs it in
* Alice does not see their existing checks and integrations, has no clue why they disappeared, and how to get them back

rophy:
Indeed.

How about make this change an option which must be explicitly toggled on? So it does not impact existing accounts.

cuu508:
Yes, I think that's what we will need to do. 

We could have a boolean setting `REMOTE_USER_HEADER_FORCE_LOWERCASE` with the default value of `False`.
If set to true, then `CustomHeaderMiddleware` converts the header value to lowercase. 

If you have an idea for a better name for the setting, I'm open to it.

Would you be able to update the PR with this (including a new test that exercises the new setting, and a new entry in `templates/docs/self_hosted_configuration.md`)?

rophy:
May be it depends on how we want to interpret the reported issue:

- Normalized REMOTE_USER_HEAD is a "special feature" that should be explicitly turned on:  `REMOTE_USER_HEADER_FORCE_LOWERCASE=False`

or

- Normalized REMOTE_USER_HEAD is a "fix", but to keep backward compatibility: `REMOTE_USER_HEADER_COMPATIBLE_MODE=True`

rophy:
Updated #1075 for `REMOTE_USER_HEADER_FORCE_LOWERCASE`.

cuu508:
[...]

PR #1075, opened 2024-10-20
Fix for #1074

rophy:
For #1074
Fix conflicts on case sensitivity between REMOTE_USER_HEADER and other places


rophy:
@cuu508 any comments on this 1L fix?

Issue #1040, opened 2024-08-02
Option to show time of last ping instead of time duration since last ping


cuu508:
### Discussed in https://github.com/healthchecks/healthchecks/discussions/1029

<div type='discussions-op-text'>

<sup>Originally posted by **StephenWetzel** July 19, 2024</sup>
On the main list of checks, in the "Last Ping" column, right now they are all relative times, like "12 hours ago".  I would also like to be able to see the absolute time the ping happened, like "2024-07-18 04:13:34".  I know I can click into it to see the times, but I'd like something easier.  Ideally, there would be a setting to allow me to toggle between them, but if you don't want add that, just showing the time when you hover over the relative time seems to be a pretty standard UI.  See for [example the Github UI itself](https://github.com/healthchecks/healthchecks/issues) where if you hover over time like "last week" it shows the timestamp.</div>

mickBoat00:
Would you like to just have the hover effect, or would you prefer the hover effect along with a setting to toggle between the formats?
Also, do you want to display the date and time in the format "2024-07-18 04:13:34" or in a more readable format like "August 24, 2024, 5:47 p.m."?

cuu508:
The approach I would like to try first is:

* instead of hover, use a tooltip (similar to the one you see when you hover over the "Name" or "Last Ping" table headers in the checks list)
* in the tooltip, show a date formatted using the browser's timezone. Use the same format as in the "Ping Details" dialog:
* 
![image](https://github.com/user-attachments/assets/d6a68ffa-1850-4aee-98b1-f081e6c92dd7)

PR #1089, opened 2024-11-25
tooltip added to display browser's timezone in ping details

NKETIAH53:


NKETIAH53:
feature for issue # 1040

cuu508:
Can you please describe the thinking behind this PR?

Looking at the code changes, IIUC, the tooltip would show the timestamp, formatted using browser's timezone, over each time cell. It would be the same date that's already displayed in the "Browser's time zone" cell. Why also display it in the tooltip?

NKETIAH53:
With respect to the earlier conversation regarding what was expected, @mickBoat00 suggested displaying the time in full on hover. 
I only based the PR on your response to his comment, considering it may be the first step in a series of approaches you would prefer to be undertaken

cuu508:
> I only based the PR on your response to his comment, considering it may be the first step in a series of approaches you would prefer to be undertaken

Did you by any chance use an LLM for this?

NKETIAH53:
No, I did not.

cuu508:
The "Ping Details" dialog is already fine – it displays the timestamp in browser's timezone in the "Browser's time zone" cell. There is no benefit to show it also in a tooltip.

Issue #1040 is about displaying a timestamp (instead of a relative value) *in the checks list*, "Last Ping" column.

PR #1094, opened 2024-11-29
Fix Trello integration (token regexp too specific)

jacekkow:
When creating Trello integration I got 400 after selecting the board to put notifications into.

It turns out the token I got from Trello does not match the regular expression as it has the following format:
"ATTA" + 64 hex chars (lowercase) + 8 hex chars (uppercase)

That "ATTA" prefix does not match `^[0-9a-fA-F]{64,256}$`. Just in case I changed the range from `a-fA-F` to `a-zA-Z` and added `_` and `-` (all base64 url-safe characters).

### DIFF
--- a/hc/front/forms.py
+++ b/hc/front/forms.py
@@ -329,7 +329,7 @@ class AddZulipForm(forms.Form):
 
 
 class AddTrelloForm(forms.Form):
-    token = forms.RegexField(regex=r"^[0-9a-fA-F]{64,256}$")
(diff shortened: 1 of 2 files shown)


jacekkow:
I'll update the pull request that way.

jacekkow:
Done. Trello token is now an opaque CharField with up to a 1000 characters.

cuu508:
Thanks, looks good!

PR #1102, opened 2024-12-11
New "Zammad integration"

christophdb:
New "Zammad" integration to create tickets directly in zammad in case of a healthcheck failure.


christophdb:
The integration is working so far. The zammad connection can be configured and tickets are created via api-endpoint of zammad.

These are three things, I couldn't solve:
- It seems that you use a sprite or a custom font for the small icons in combination with css. I used a "wrong" content for .ic-zammad:after and .ic-zammad:before. 
- I disabled "edit" the configured zammad integration. I couldn't figure out how to allow editing. It always returned an error page if I clicked on edit.
- now for "up" and "down" a new ticket is created in zammad. it would be better if the "up" event creates an update of the existing ticket instead of creating a new ticket. I couldn't figure out to save/use the ticket id for the "up" event.

christophdb:
Here some screenshots:

New available integration:
image

Konfiguration page:
image

Required personal access token from Zammad:
image

Ticket in Zammad:
image

cuu508:
Thanks for the PR!

I didn't know about Zammad when I first saw this PR, and so was putting off reviewing the PR. After looking into it, happy to work with you to get this merged.

> It seems that you use a sprite or a custom font for the small icons in combination with css. I used a "wrong" content for .ic-zammad:after and .ic-zammad:before.

No worries, I can take care of this. Editing the icon font is fiddly, easier to do it myself than to document the whole process ;-)

> I disabled "edit" the configured zammad integration. I couldn't figure out how to allow editing. It always returned an error page when I clicked on edit.

Editing is useful for integrations where you would sometimes want to tweak the parameters a couple times until you get them right. Webhooks are the prime example. For integrations like Zammad that use predetermined payload I think it's fine not to have the Edit button. If the personal access token changes, the user can delete the integration and create a new one.

> now for "up" and "down" events a new ticket is created in zammad. it would be better if the "up" event creates an update of the existing ticket instead of creating a new ticket. I couldn't figure out to save/use the ticket id for the "up" event.

Yes, we don't have a place on Healthchecks side to stash the ticket id into, for using it later. 

Looks like Zammad has ticket search: https://docs.zammad.org/en/latest/api/intro.html#endpoint-search
Can you try the following:

* When creating the ticket, include the check's code (UUID) in the ticket body
* On the "up" event run a search API call to look up the most recent ticket with a matching UUID, then update it (and do nothing if the search returns no matching ticket).

---

Could you also add some tests?

* A test that exercises the "Add Zammad" form
* A test that exercises the Zammad transport class in transports.py

cuu508:
No response, closing.

PR #1104, opened 2024-12-15
Set pipefail true for bash in Dockerfile

aminvakil:


cuu508:
We use pipe in a single place, in the line that installs rust. How about setting pipefail just there? Something like:

```
RUN set -o pipefail && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | bash -s -- --profile minimal -y
```

I'm suggesting this because docker's [Best Practices](https://docs.docker.com/build/building/best-practices/) docs has this example. I don't know about pros and cons of either approach.

cuu508:
Ah, but the base image uses dash as the default shell, which does not support pipefail. So this would have to be

```
RUN ["/bin/bash", "-c", "set -o pipefail && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | bash -s -- --profile minimal -y"]
```

cuu508:
If we go with the SHELL approach, doesn't the command need a `-c` parameter at the end?

```
SHELL ["/bin/bash", "-o", "pipefail", "-c"]
```

aminvakil:
> We use pipe in a single place, in the line that installs rust. How about setting pipefail just there? Something like:
> 
> ```
> RUN set -o pipefail && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | bash -s -- --profile minimal -y
> ```
> 
> I'm suggesting this because docker's [Best Practices](https://docs.docker.com/build/building/best-practices/) docs has this example. I don't know about pros and cons of either approach.

Yes, that crossed my mind first, but I'd thought maybe adding it to Dockerfile would remove the necessity for adding it for further usages, if there would be any.

But I would change it if you think this wouldn't be necessary later.

aminvakil:
> If we go with the SHELL approach, doesn't the command need a `-c` parameter at the end?
> 
> ```
> SHELL ["/bin/bash", "-o", "pipefail", "-c"]
> ```

Yes, it needs -c at the end.

cuu508:
Thanks. Let's go with the `SHELL` approach.

PR #1116, opened 2025-01-21
Fix allow rid variable to be upper and lowercase

wie-niet:
Shell scripts with healthchecks to measure execution time using 'rid' were failing on my MacOS (they were created in uppercase instead of lowercase). 

With this patch healthcheck now supports 'rid' created with /usr/bin/uuidgen on MacOS. 
The uuid.UUID module in python interpret lower and upper case uuids , only the validation in 'hc/lib/string.py' was restricted to only lowercase.

Now it supports both ;-)

### DIFF
--- a/hc/lib/string.py
+++ b/hc/lib/string.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import re
 
 uuid_match_regex = re.compile(
-    "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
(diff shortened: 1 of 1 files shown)


cuu508:
Thanks for the PR, merged it here: https://github.com/healthchecks/healthchecks/commit/3162ef64848ebc2022f9995b695f8dc3f25e08f2

On one hand, I'd like to keep things consistent and accept a single, specific UUID format everywhere. On the other hand, it's an easy fix and makes life a little easier for MacOS users.  In this case, for me, user convenience trumps purity and pedantry so this goes in.

PR #1146, opened 2025-04-11
Fix MariaDB version check in mariadb_uuid_check()

fcharlier:
The version check implemented was wrong for e.g. version "11.3".

This commit makes the version check more robust.

### DIFF
--- a/hc/api/apps.py
+++ b/hc/api/apps.py
@@ -162,7 +162,7 @@ def mariadb_uuid_check(
         # - we are using MariaDB 10.7+
         # - *and* the UUID columns exist and use a varchar datatype,
         # then we have a problem.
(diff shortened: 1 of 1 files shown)


cuu508:
Thanks and sorry – what an embarrassing bug by me 😬

The `packaging` module happens to be installed on my system, but I'm not sure we can assume it will be installed in the general case. I just tested in the docker image, and it is absent there. We could add it to the requirements of course, but perhaps a simpler fix would be to use tuple comparison:

    if (major, minor) >= (10, 7) and data_type == "char":

fcharlier:
@cuu508 that's much simpler with tuple comparison indeed.

I updated the change.

cuu508:
That's great, thanks!

PR #1165, opened 2025-05-25
Add prefixes to API keys. Closes #1148

### LINKED ISSUE #1148: Use human-readable API key prefixes to distinguish read-only and read-write API keys
Hi!

It has come to my attention that there is a (good) warning "Be sure to use the read-only API keys!" in the readme of https://github.com/healthchecks/dashboard and that made me remember that the problem could be prevented by design and altogether by using slightly different syntax for these two kinds, e.g. the read-only ones always starting with `R` and the writable ones starting with `W`. That would allow the backend to detect and reject token mix-up safely and easily.

The idea is from article https://blog.hboeck.de/archives/909-Mixing-up-Public-and-Private-Keys-in-OpenID-Connect-deployments.html where the consequences of mix-up were even more severe.

Thanks for your consideration!

CC @hannob

lobovkin:
- Added "RW_" prefix to write-enabled API keys for clear identification
- Added "RO_" prefix to read-only API keys
- Fixed test_it_creates_readonly_key

### DIFF
--- a/hc/accounts/tests/test_project.py
+++ b/hc/accounts/tests/test_project.py
@@ -65,8 +65,8 @@ class ProjectTestCase(BaseTestCase):
         self.assertEqual(r.status_code, 200)
 
         self.project.refresh_from_db()
-        self.assertTrue(self.project.api_key_readonly.startswith('RO_'))
-        self.assertEqual(len(self.project.api_key_readonly), 3 + 32)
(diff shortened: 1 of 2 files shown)


lobovkin:
I simplified the **_new_key** function, and all of the above requirements have been met.

PR #1173, opened 2025-06-04
GITHUB_CLIENT_ID typo in settings.py

lobovkin:
There was a copy-paste error in this line:
GITHUB_CLIENT_ID = os.getenv("LINENOTIFY_CLIENT_ID")


cuu508:
Good catch, thanks!

PR #1177, opened 2025-06-12
docker: default to listening on IPv6 as well

youknow0:
When only a port is specified, uWSGI will only listen on the IPv4 wildcard (0.0.0.0) address. This will break environments relying on IPv6.

Use `http-socket = [::]:8000` to have uWSGI listen on IPv4 as well as IPv6.

### DIFF
--- a/docker/uwsgi.ini
+++ b/docker/uwsgi.ini
@@ -2,7 +2,7 @@
 strict
 master
 die-on-term
-http-socket = :8000
+http-socket = [::]:8000
 harakiri = 10
 buffer-size = 32768
 post-buffering = 16192


cuu508:
So I understand the problem, can you give an realistic example of such an environment?

The healthchecks instance should be behind a reverse proxy, is the reverse proxy running on a different machine, and the healthchecks instance is running on a machine without a public IPv4 address?

youknow0:
Yes, that would be one example. Another one would be if the docker network is configured to be ipv6-only.

cuu508:
>  Another one would be if the docker network is configured to be ipv6-only.

Is this also your case? Is there any particular benefit of configuring docker network as ipv6-only? Do people often do that and why? And do you know of other Docker images that listen on [::]:<port> for this specific reason?

I'm asking because I'm familiar with Docker and IPv6 only at the a surface level, and I want to understand the ins and outs of the problem and the pros and cons of the solution before I hit merge.

youknow0:
Ah, did you run `netstat` on the host or inside the container?

On the host, it's docker's proxy listening. It will proxy IPv6 to IPv6 and IPv4 to IPv4. Hence, when I explicitly connect via IPv6 from the host, it won't work with 3.10:
```
# curl -v http://[::1]:8080
*   Trying [::1]:8080...
* Connected to ::1 (::1) port 8080 (#0)
> GET / HTTP/1.1
> Host: [::1]:8080
> User-Agent: curl/7.88.1
> Accept: */*
> 
* Recv failure: Connection reset by peer
* Closing connection 0
curl: (56) Recv failure: Connection reset by peer
```
while IPv4 works.

When I use 3.11, IPv6 also works from the host.

PR #1182, opened 2025-06-18
Update README.md Telegram guide

0x412e4e:
Added information that the settelegramwebhook command needs to be run inside the web container.


cuu508:
Thanks for the PR. I understand the confusion you had, and appreciate the attempt to improve the docs, but I don't think this is the right way.

The README is written from the perspective of setting up a Healthchecks instance on a host machine, inside a virtualenv.  You can run Healthchecks fine without using containers. Adding docker-specific instructions in one section would make things confusing for people following instructions in README step-by-step.

We could perhaps add a "Running manage.py commands" section in the README that's inside the /docker/ directory.

PR #1199, opened 2025-08-16
Use From header instead of envelope sender in SMTP ping source

purplebite:
## Problem
Currently, the SMTP ping feature shows the envelope sender (Return-Path) as the source, which often displays technical bounce addresses instead of the actual sender's email address.

## Solution
This PR modifies the `_process_message` function in `smtpd.py` to:
- Parse the email message and extract the `From` header
- Use the `From` header value in the user agent string instead of the envelope sender
- Fall back to the envelope sender if no `From` header is present

## Changes
- Modified `_process_message()` to parse email message earlier
- Extract `From` header using `message.get("From", mailfrom)`
- Update user agent string to use `From` header value

## Testing
- [ ] Existing tests should pass (logic unchanged for core functionality)
- [x] Maintains backward compatibility  
- [x] Handles missing From header gracefully (fallback to envelope sender)
- [ ] Needs testing with real SMTP messages

**Note**: This change follows existing patterns in the codebase but would benefit from testing with actual email messages.

### DIFF
--- a/hc/api/management/commands/smtpd.py
+++ b/hc/api/management/commands/smtpd.py
@@ -82,12 +82,16 @@ def _process_message(remote_addr: str, mailfrom: str, mailto: str, data: bytes) 
         except Check.DoesNotExist:
             return f"Check not found: {mailto}"
 
+    # Parse email message to extract From header
+    data_str = data.decode(errors="replace")
+    message = email.message_from_string(data_str, policy=email.policy.SMTP)
+    assert isinstance(message, EmailMessage)
+    
+    # Extract From header, fallback to envelope sender
+    from_header = message.get("From", mailfrom)
+
     action = "success"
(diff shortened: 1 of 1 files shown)


cuu508:
Can you please confirm this [was not produced with the help of AI tools](https://github.com/healthchecks/healthchecks/blob/master/CONTRIBUTING.md#ai-policy)?

purplebite:
I was not aware of your (very) strict AI rule. But yes, of course I had help from AI: not only to write a clear PR summary and learn proper contribution practices, as this is my first contribution to an existing open source project, but also to grasp Python, as the last real coding I did was when Perl was a dominant language.

Is _any_ help therefore not quite strict? Feel free to close the PR, even if I believe this technical improvement has merit and lots of human research was done for it.

PR #1225, opened 2025-10-31
Use psycopg[c] and removed specific version

solracsf:
To be in line with changes on `requirements.txt` file
and always install a `c` version matching the `psycopg` version .


cuu508:
Thanks!

